Issue #1123, opened 2015-12-24
getting equipment ids


rs-sliske:
after some discussion / testing in the dev skype chat, we found the ids returned from appearance are offset (so the game can tell what player looks like when they arent wearing anything)

would be nice to be able to get the item id because that is going to be more useful

wrote a small method which cancels out the offset but would be nice to have something in the api

```java
public int getEquipmentID(final Player player,final int index){
	return player.appearance()[index] & 0xfffffdff;
}
```

rs-sliske:
```java
public int getEquipmentID(final Player player,final int index){
	return player.appearance()[index] & 0xfffffdff;
}
```

Timer:
What game version is this for?

rs-sliske:
Osrs, might be useful in darkscape aswell but i havent tried it there

rs-sliske:
https://github.com/powerbot/powerbot/issues/1123

Timer:
See a70ea62278b309e00728d3ff05409c449fc4d422.

PR #1128, opened 2015-12-27
get equipment ids from player

rs-sliske:
https://github.com/powerbot/powerbot/issues/1123


Timer:
These methods are misplaced and please add symmetrical ones to rt6 if you wish for them to be implemented; however, I have brought the id fix into upstream.
Please re-open if you feel this is still necessary.

Issue #1110, opened 2015-12-17
Item instance exists, but has no values


Moosheplusplus:
Reporting on behalf of [Laputa's issue on powerbot] (http://www.powerbot.org/community/topic/1293771-client-booleans-not-working-properly/) - I was able to replicate the bug.

```java
ItemQuery<Item> c = ctx.inventory.select().action("Eat");
if(c.isEmpty())
	log.info("No food");
else
	for(Item i : c)
		log.info("Item found: "+i.name());
```

Trying to select all inventory items with the action "Eat" when there's only Monkfish will return an empty ItemQuery.

Output of above code with just Monkfish:
> [INFO] pw.mooshe.merch.MainScript: No food

Output of above code with both Monkfish and Shrimp:
> [INFO] pw.mooshe.merch.MainScript: Item found: Monkfish
> [INFO] pw.mooshe.merch.MainScript: Item found: Shrimps

Output of above code with just Shrimp:
> [INFO] pw.mooshe.merch.MainScript: Item found: Shrimps

This is only reproducible if there are no different types of items with the action "Eat". If there is another item with "Eat", it will return both normally.

```java
ItemQuery<Item> c = ctx.inventory.select();
for(Item i : c)
	log.info("Actions for "+i.getName()+": "+Arrays.asList(i.actions()));
```

When I try to output the actions of each item in the inventory with Shrimp and Monkfish in the inventory, it produces this result:
> [INFO] pw.mooshe.merch.MainScript: Actions for Monkfish: [Eat, null, null, null, Drop]
> [INFO] pw.mooshe.merch.MainScript: Actions for Shrimps: [Eat, null, null, null, Drop] 

If I have just Monkfish in the inventory, the ItemQuery has an item in it, but it has no values whatsoever (no name, no item ID, no actions, etc.)

It doesn't seem like Monkfish registers properly with RSBot if it's the only food item in the inventory.

AnthonyOSX:
Possible duplicate and/or related to #1109

Timer:
This is because we do not yet have cache definitions for rt4 items, perhaps @iitc can help with this.

iitc:
@Timer, item definitions for OS are identical to RS3 but for opcodes
```
                case 1:
                case 23:
                case 24:
                case 25:
                case 26:
                case 78:
                case 79:
                case 90:
                case 91:
                case 92:
                case 93:
```
you read a unsigned short rather than the unsigned "packed int" that chooses between a short/int.

[...]

PR #1143, opened 2015-12-28
Added RT4 item definition from cache

Logicail:
Can you check Item.name()

RT6 can use Component#itemName() but that doesn't exist for RT4.
I'm not sure if using itemConfig will always work or if it should just always load from the cache.

Tested the class loading items 0-13352

### DIFF
--- a/rt4/CacheItemConfig.java
+++ b/rt4/CacheItemConfig.java
@@ -1,0 +1,195 @@
+package org.powerbot.script.rt4;
+
+import org.powerbot.bot.cache.Block;
+import org.powerbot.bot.cache.CacheWorker;
+import org.powerbot.bot.cache.JagexStream;
+
+class CacheItemConfig {
+    public final int index;
+    private final CacheWorker worker;
+    private final JagexStream stream;
+    public String name = "";
+    public boolean tradeable;
+    public boolean stackable;
+    public boolean members;
+    public int team = -1;
+    public int value = -1;
+    public int modelId = -1;
+    public int modelZoom = 0;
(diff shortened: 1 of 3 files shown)


rt4/Item.java
```diff
@@ -53,7 +54,14 @@ public Point centerPoint() {
 
 	@Override
 	public String name() {
-		return StringUtils.stripHtml(ItemConfig.getDef(ctx, id).getName());
+		final String name;
+		if (component != null && component.itemId() == id) {
+			name = ItemConfig.getDef(ctx, id).getName();
```
Logicail: Not sure about this RT6 uses Component#itemName()

rt4/CacheItemConfig.java
```diff
@@ -0,0 +1,195 @@
...
+                    modelOffsetY -= 65536;
+                }
+            } else if (opcode == 11) {
+                stackable = true;
+            } else if (opcode == 12) {
+                value = stream.getInt();
+            } else if (opcode == 16) {
+                members = true;
+            } else if (opcode == 23) {
+                final int i = stream.getUInt24();
```
Logicail: 23 & 25 actually read ushort and ubyte

Logicail:
#1110

rt4/CacheItemConfig.java
Timer: respectively or both in each?
you changing this or want me to go through it all?

Logicail: Both in each, the value isn't used anyway and reading 2+1 byte is the same as 3 bytes.

PR #1155, opened 2015-12-29
Add scripts

xXTheAwesomerXx:


ChrisBPB:
You shouldn't be adding all those git ignores

xXTheAwesomerXx:
Just realized that I had committed those, will revert and recommit

PR #1157, opened 2015-12-29
Optimize Objects select

Moosheplusplus:
Optimize Objects select to define the local point, so instead of setting the radius to the distance of the tile and player, you can select objects around the tile.


iitc:
`public List<GameObject> get(final Tile at, int radius)` should be `public List<GameObject> get(final Locatable locatable, int radius)` if this gets added.

Timer:
@Moosheplusplus: merged with 181788d9433afa94ae35ce43178cbc94c67375b6.
@Overflow-Powerbot Valid optimization in cb091de859bb18285d86939dfb8607478ce69b36.

PR #1158, opened 2015-12-29
Updated rt4 items

ArtificialPB:
Added modelId() method which is useful for filtering components as some have item modelIds instead of item ids. Also updated loading noted items from template id to prevent NPEs.


Timer:
For future reference, please see the codeStyleSettings in the repo.

Timer:
Merged with f1356425809551c38e1b9fa5e4e7d37754ce9be4

ArtificialPB:
Will do.

PR #1169, opened 2015-12-31
Switching to Yews

xXTheAwesomerXx:
++ Support for switching at Yews at level 60
~ Better chopping methods - now turns to next tree in the vicinity


Strikegently:
Merged.

Issue #1173, opened 2015-12-31
Could Area#getTiles be made public?


SD8Z:

AnthonyOSX:
Why don't you do it yourself and create a pull request?

Timer:
Why do you need said method?

SD8Z:
Nothing too fancy. I personally would just use it to draw the Area. Back in RSBot 2 there was a method that would draw the area but that is no longer here.
By having access to the array of tiles, a person could be able to get the TileMatrix and therefore draw each tile, which would be drawing the area.
Obviously this is just for aesthetic purposes, and (at least me) wouldn't draw an area bigger than 3x3.

Timer:
@SD8Z the problem is we have to consider the viability of such a method and its memory implications when someone uses an area larger than 3x3. Can you not use the tiles you generated the area with? Or get the polygon and compute each point, constructing a custom polygon and drawing.

PR #1176, opened 2015-12-31
Update Area.java

SD8Z:
Made getTiles() public.


Timer:
What is this necessary for?

Timer:
Please discuss this in the relevant issue, #1173. This pull request is denied because it is providing raw access to an internal array which may be misused by the scripter and cause corruption.

PR #1179, opened 2016-01-01
Avoids trying to open equipment tab if bank is open

ArtificialPB:
Title


ArtificialPB:
Fixed an issue with Hud#opened() in legacy mode where it would try to open the LegacyTab instead of just checking if it's opened.

Timer:
Nice catch, @ArtificialPB

Timer:
@ArtificialPB can you please rebase and remove your edits to the imports?

ArtificialPB:
@Timer Fixed imports

Timer:
Thanks!

Issue #1178, opened 2016-01-01
Parse equipment without opening tab


ArtificialPB:
Is it possible to parse equipped items without having to open equipment tab in rt4?

rs-sliske:
Ctx.players.local.appearance

ArtificialPB:
It returns wrong ids for a lot of items @rs-sliske

Timer:
@ArtificialPB the appearance was fixed, but I can look into without opening tab (or did you fix that in your PR @ArtificialPB)?

ArtificialPB:
I did not fix that in my pull request @Timer

ArtificialPB:
Already merged, closing this.

PR #1181, opened 2016-01-01
Fix player appearance offset

Logicail:
There have been a few commits to this recently, but it was still broken.

Code to demonstrate that using a mask doesn't work #1178 (latest change isn't in client build yet)

```
int wrong = 0, correct = 0;
for (int i = 0; i < 13353; i++) {
    if (((i + 512) & 0xfffffdff) != i) {
        //System.out.println(i + " wrong returns " + ((i + 512) & 0xfffffdff));
        wrong++;
    } else {
        correct++;
    }
}
System.out.println("wrong: " + wrong + ", correct:" + correct);
```

It will currently only work in about half cases (not exactly since you can't wear all items)

`wrong: 6656, correct:6697`

If anyone wants to check this is a list of wearable items that I think will give the incorrect id with the current method https://gist.github.com/Logicail/1efbb5b2b74afe14bec2

### DIFF
--- a/rt4/Player.java
+++ b/rt4/Player.java
@@ -47,7 +47,7 @@ public class Player extends Actor {
 		}
 		arr = arr.clone();
 		for (int index = 0; index < arr.length; ++index) {
-			arr[index] = arr[index] < 512 ? -1 : arr[index] & 0xfffffdff;
+			arr[index] = arr[index] < 512 ? -1 : arr[index] - 512;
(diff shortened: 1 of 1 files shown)


rs-sliske:
fk me, i guess i just got lucky the things i tested worked

ill stay away from bitwise stuff from now on though unless im 100% sure it works for every possible scenario

PR #1186, opened 2016-01-02
Fixed Import

xXTheAwesomerXx:
Import was incorrectly packaged, removed.


Strikegently:
You should really test that your scripts compile before committing your changes.

xXTheAwesomerXx:
My script did compile locally because that specific package was located in my project. The issue was that I did not upload the supporting package itself.

PR #1194, opened 2016-01-05
RT4 Equipment parsing

ArtificialPB:
Added https://github.com/powerbot/powerbot/issues/1178. If the equipment tab is not opened it will set item stack size to 1.

@Timer is it possible to parse stack sizes like in rt6? https://github.com/powerbot/powerbot/blob/rsbot-api/rt6/Equipment.java#L20

### DIFF
--- a/rt4/Equipment.java
+++ b/rt4/Equipment.java
@@ -18,15 +18,14 @@ public class Equipment extends ItemQuery<Item> {
 	@Override
 	protected List<Item> get() {
 		final List<Item> items = new ArrayList<Item>(11);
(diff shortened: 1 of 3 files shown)


Timer:
@ArtificialPB I'll have to see if this hook exists in rt4, I have a feeling it doesn't.

ArtificialPB:
Updated itemAt(Slot) to account for new equipped items parsing, fixed small widgets issue, updated minimap parsing for Game#tileToMap(Tile) to it does not brake every other game update.

Timer:
@ArtificialPB can you get on slack?

PR #1195, opened 2016-01-05
DBWoodcutter - Add Ivy

xXTheAwesomerXx:
++ Added Falador SE and Varrock N with support for Ivy!
++ Added secondary location for Willows
++ Added support for banking from secondary willows, oaks, and regular trees
~ Script now loads best methods from 1-99 by default


Strikegently:
Merged.

Issue #1196, opened 2016-01-05
Bank pin not being entered


ChrisBPB:
Also bank.opened isn't returning true when bank pin screen is open (I think it should be?)

Main widget is 213, if needed then 213,7 contains the text "enter your pin"

Widgets for the numbers start at 213,16 and go every other one (i.e 213,16   213,18   213,20 ... 213,34, the value is found at 213,X,1

Timer:
I need an account with a bank pin!

Timer:
still need account with bank pin

ChrisBPB:
I have a pin set now, let me know on Skype when you want it.

ChrisBPB:
still an issue

Timer:
added

Issue #825, opened 2015-02-27
Bank#open and Bank#nearest


iitc:
Bank for rt4 needs opening methods. 
If you want a code reference, https://github.com/IITC/Powerbot/blob/master/src/iitc/rsbot/script/rt4/Bank.java
It's messy at the moment due to #818. I also support isolating individual bank types as interacting with bankers sometimes leaves you outside the bank; something I liked to combat.

Strikegently:
This is already done for v7. 

If you remember, the reason it wasn't available before was because of the constantly-changing IDs.

IITC:
Well if it's done for v7, sweet. Bank type isolation is a necessity/ Interacting with bank booths alone relieves the issue of getting stuck outside banks.

Timer:
Added in #1201

PR #1201, opened 2016-01-08
RT4 prayers and bank update

ArtificialPB:
Finished prayers, moved Prayer constants to Constants class, added varpbit helper methods.

### DIFF
--- a/rt4/Prayer.java
+++ b/rt4/Prayer.java
@@ -1,5 +1,7 @@
 package org.powerbot.script.rt4;
 
+import java.util.LinkedHashSet;
+import java.util.Set;
 import java.util.concurrent.Callable;
 
 import org.powerbot.script.Condition;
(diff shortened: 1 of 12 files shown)


ArtificialPB:
Added Bank#open(), #nearest(), #present() and #inViewport(), added new Nameable.Query helper method for string arrays. Only thing missing for bank is https://github.com/powerbot/powerbot/issues/1196.

This also completes https://github.com/powerbot/powerbot/issues/825 request with the exception of bank type isolation.

rt4/Prayer.java
```diff
@@ -13,24 +15,193 @@ public Prayer(final ClientContext ctx) {
...
+				return quickSelectionActive() == quick;
+			}
+		}, 150, 10);
+	}
+
+	public boolean prayerQuick(final Effect effect) {
+		return ctx.varpbits.varpbit(Constants.PRAYER_QUICK_SELECTION, effect.ordinal(), 0x1) == 1;
+	}
+
+	public Effect[] activePrayers() {
```
AnthonyOSX: I'd recommend returning a list on these methods.

rt4/Prayer.java
```diff
@@ -13,24 +15,193 @@ public Prayer(final ClientContext ctx) {
...
+		final Set<Effect> active = new LinkedHashSet<Effect>();
```
AnthonyOSX: Why not use a LinkedList?

ArtificialPB: A Set if faster than a List for additions.

Logicail:
Are the breaking changes to method names necessary?

Timer:
Leaving this PR open for discussion for a while.
And @Logicail is correct, you cannot break api. And `ctx.prayer.prayerPoints()` is redundant, what was wrong with `ctx.prayer.points()`?

rt4/Prayer.java
ArtificialPB: I converted these methods from rt6 and did not change return types. I can change them to Lists though.

Timer: We should keep it consistent between API versions.

rt4/Prayer.java
Timer: Set and lists are both O(1).

Logicail:
Also could you please make separate branches for PRs so all your changes don't end up in one PR

ArtificialPB:
@Timer I changed it to match rt6 method names from Powers class.
@Logicail Sure I'll make a separate pull requests for future commits.

rt4/Prayer.java
Strikeskids: And LinkedHashSet has additional overhead from the dual chaining.

Strikeskids:
Also, there is a lot of mixing between final and non-final arguments. It gives false significance to the non-final ones.

rt4/Prayer.java
ArtificialPB: I've seen a benchmark between different types of collections and Set was faster for additions than List. I'm aware that benchmarks are not 100% accurate. I just converted those methods from rt6. As far as I'm concerned, it can be changed to a List.

ArtificialPB:
@Strikeskids Most of the code was converted from its rt6 versions, with the exception of Prayer.Effect enum and new Nameable.Query helper method. I will update those arguments.

rt4/Prayer.java
AnthonyOSX: Most benchmarks aren't performed correctly. Debug builds always have additional instructions to help with debugging, the garbage collector could kick in, JIT warm up, etc.

ArtificialPB:
Ok, I've removed breaking changes and deprecated old methods in Prayer class and made some non-final arguments final. 
I think I messed something up when modifying commits and pushing them so let me know if I need to change anything @Timer

iitc:
Returning an array for active/quick prayers makes it almost useless. I'd recommend using a Set for lookup purposes but that most likely won't match the current API design. Methods for checking if a prayer is selected for quick/normal use should be provided. Also, flattening a 2d array into a 1d array can be O(n) where n is the number of 1d arrays if you make proper use of array copies and just sum up the lengths of the arrays for the new one.

Strikeskids:
@iitc flattening an array can never be O(number of arrays). You are making a new array with many elements, so at best it can be O(total size of all arrays).

ArtificialPB:
@iitc This does have a method do determine if a prayer is selected as quick prayer: https://github.com/ArtificialPB/powerbot/blob/rsbot-api/rt4/Prayer.java#L92-L94

iitc:
@Strikeskids, that's if you're doing manual copies where you're copying index to index. System array copies will drastically speed this up.

Strikeskids:
@iitc just because it has a lower constant doesn't mean it's asymptotically faster. If you have 10 arrays of size 30 and 10 arrays of size 2 million, the second flatten is going to be slower.

iitc:
@Strikeskids, The implementation of system array copies is O(1). I don't know how that'll reflect on the JVM level but the native implementation of the algorithm is what I'm referencing.

Strikeskids:
@iitc I think you are misinformed about runtime asymptotic complexity. Although system array copy (using memmove or memcpy) performs better, because it can leverage properties of the hardware, it is still a linear time algorithm. For small array lengths, you can think of it having EFFECTIVELY constant time; however, that thought process breaks down when the array length increases substantially.

If you have an array with 5 million elements and an array with 10 million elements, the one with 10 million will take twice as long to copy: O(n).

iitc:
@Strikeskids, I'm well aware of asymptotic complexity. For some reason I thought that you could transfer blocks of data in a single operation but I was just not thinking. Regardless, making use of system array copy and just flattening the array is FAR more efficient from a speed and memory standpoint than using sets. Using an ArrayList with a preset capacity would function similarly as it's just a wrapper and is up to @Timer to choose. I believe he just manually flattens the arrays without the use of a list.

Timer:
@ArtificialPB can you rebase this please?

ArtificialPB:
Why do I have the feeling something went wrong here? lol I'm going to solve it in the morning.

Timer:
looks very wrong, that's why i asked for rebase in the first place, it shows that i authored this lol @ArtificialPB

ArtificialPB:
I solved some issues with yesterdays push, but I have no idea how to change the author of this commit lol Hit me up when you get on Slack @Timer

Strikegently:
@ArtificialPB 
it should just be `git rebase upstream/rsbot-api`

ArtificialPB:
Yeah I already did that and it says "Current branch rsbot-api is up to date." @Strikegently

PR #1202, opened 2016-01-09
Elder update V2.02 - Randomization and teleport fixes

helicopter12:
Reverted to old randomizing method as the new method was said to have issues with locked lodestones. Also added checks to the teleporting steps to ensure the widgets are clicked before proceeding. Otherwise, the script will just sit there.


Strikegently:
Merged.

PR #1203, opened 2016-01-09
Elder: Fix new randomization. Store respawn times for all trees

Strikeskids:
Previous bug was double-indexing. Found through user testing.

Storing all respawn times preserves more information, to make the script smarter when doing antipattern.


Timer:
@Strikegently this was approved again by helicopter12

PR #1207, opened 2016-01-10
Elder update V2.03 - Out of Bound Exception Fix

helicopter12:
*Fixed out of bounds exception that caused the script to stop running after it cycled once through all 4 locations.
*Added XP tracking to the paint and adjusted the paint to accommodate the new line of paint. 
*Added jujupotion ID for future reference
*Lowered anti-ban execution rate


Strikegently:
Merged.

PR #1214, opened 2016-01-16
Moved map point checking from TileMatrix#onMap() to Game#tileToMap()

ArtificialPB:
So you don't have to fix map components in two places when they change. This is rt4.


Timer:
Some scripts expect map point to work out of map bounds.

PR #1219, opened 2016-01-16
Added Chicken Slayer Script

awbasham:
First script on RSBot


ChrisBPB:
You should be using the Condition class instead of this TimeUnit, Condition.wait also randomizes the amount so you won't need Random.netInt etc

awbasham:
@ChrisBPB What's the benefit of using Condition over TimeUnit?  Only the removal of Random.nextInt?  I was using what I was familiar with, of course.  Can change if needed.

ChrisBPB:
The condition class can offer you a lot, in addition to all sleeps being "randomly adjusted to provide a basic antipattern." there are other methods available https://www.powerbot.org/rsbot/docs/org/powerbot/script/Condition.html specifically the ever popular 'dynamic sleeps' where you can sleep until a boolean returns true

The following snippet for example will sleep until it times out (tries every 1000ms, maximum of 3 times) or until the boolean returns true
```java
Condition.wait(new Callable<Boolean>() {
    @Override
    public Boolean call() throws Exception {
        return obs.inViewport();
    }
}, 1000, 3);

awbasham:
@ChrisBPB Ah, thanks!  I was unaware of the condition class.  Just another issue with learning a new API..  I'll make adjustments.  This was my first script I wrote within a day of viewing the api and some forum posts.

awbasham:
@ChrisBPB As for testing, if this is added to the SDN will I be able to test for longer than 2 hours?  It runs good for 2 hours but I have no way to test for long sessions.

Strikegently:
Merged.

rs-sliske:
@awbasham if you run the sdn version, you will be able to test for 6 hours, or as long as you want if you purchase vip

your local version will still only run for 2 hours at a time

PR #1221, opened 2016-01-17
Added support for bottom line tabs rt4

ArtificialPB:
There is a different widget for tabs when in resizable mode with bottom line tabs option selected:
resizabletabs


Timer:
Thanks.

Issue #1177, opened 2015-12-31
Add new bank methods introduced to rt4 to rt6


Timer:

Strikegently:
You may use this for RT4 or RT6. It is much more elegant (in my opinion) than what is there currently. 

https://gist.github.com/Strikegently/5437d13796684e5ca1b4

Timer:
done in #1224

PR #1224, opened 2016-01-17
Added new deposit helper methods to rt6 bank, matched rt4 version

ArtificialPB:
https://github.com/powerbot/powerbot/issues/1177
Used the snippet @Strikegently provided with additional shuffle() method call before depositing.


Timer:
Thanks.

PR #1226, opened 2016-01-18
Fally Agility: v2.0

achung205:
reduced script size considerably with new logic 
checks energy level and toggles (to be tested)


Strikegently:
Be a tad more concise than that.

achung205:
yes sir.

PR #1228, opened 2016-01-19
Al-Kharid Slayer: Reduced Pattern-causing Behavior

awbasham:


awbasham:
And greatly improved door handling.

Strikegently:
Merged.

PR #1229, opened 2016-01-19
removed extraneous files

achung205:


achung205:
script not showing up on sdn

Strikegently:
You do know that this commit removes your script, right?

achung205:
i have a folder named Agility and my source files are in there. i removed the files from the andyroo/ directory. does that not work? what if i want to make another script?

Strikegently:
@achung205 
I didn't see that. All I saw was that it removed your script.

Merged.

PR #1232, opened 2016-01-19
renamed class FaladorAgility2 to FaladorAgility

achung205:
should appear on sdn now


Strikegently:
@achung205 
FYI: the SDN identifies scripts based on class name, so when you rename your main class, everyone that has your script added with lose it. They can re-add it, but sometimes it's annoying for the user.

Merged.

achung205:
yes, it was a mistake because i was juggling two versions of my script

PR #1235, opened 2016-01-20
Al-Kharid Slayer slight combat improvements

awbasham:


Strikegently:
Instead of `TimeUnit.sleep()`, you can use `Condition.sleep()`, which handles the `InterruptedException` for you (so you don't need to wrap tons of blocks of code in try-catch).

You should be using `Condition.wait()`, though, for conditional sleeping instead of just static waiting periods.

PR #1236, opened 2016-01-20
Finished rt4 Chat

ArtificialPB:
Added ChatOption class and TextQuery to rt4, deprecated some chat methods and added some additional methods that match rt6 version of Chat class.

### DIFF
--- a/rt4/Chat.java
+++ b/rt4/Chat.java
@@ -2,7 +2,6 @@ package org.powerbot.script.rt4;
 
 import java.awt.Graphics;
 import java.util.ArrayList;
-import java.util.Arrays;
 import java.util.List;
 import java.util.concurrent.atomic.AtomicBoolean;
 import java.util.concurrent.atomic.AtomicReference;
@@ -13,6 +12,7 @@ import org.powerbot.bot.rt4.client.Client;
 import org.powerbot.bot.rt4.client.Entry;
 import org.powerbot.bot.rt4.client.EntryList;
 import org.powerbot.bot.rt4.client.MessageEntry;
+import org.powerbot.script.Condition;
 import org.powerbot.script.MessageEvent;
 import org.powerbot.script.PaintListener;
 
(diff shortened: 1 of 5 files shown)


rt4/Chat.java
```diff
@@ -83,28 +118,68 @@ public boolean continueChat(final boolean useKeys, final String... options) {
...
+	}
+
+	/**
+	 * Continues the chat.
+	 *
+	 * @param key <tt>true</tt> to press space; <tt>false</tt> to use the mouse.
+	 * @return <tt>true</tt> if the chat was continued; otherwise <tt>false</tt>
+	 */
+	public boolean clickContinue(final boolean key) {
+		final Component c = getContinue();
```
Strikegently: This can be simplified a ton. 
`return c != null && (key && ctx.input.send("{VK_SPACE}")) || c.click()`

ArtificialPB: I agree, it could be simplified. This method was just copied from rt6. I'll improve this and some other things I've noticed.

rt4/Chat.java
```diff
@@ -83,28 +118,68 @@ public boolean continueChat(final boolean useKeys, final String... options) {
...
-		if (options != null) {
-			final List<String> o = Arrays.asList(options);
-			final List<Component> ol = chatOptions();
-			for (int i = 0; i < ol.size(); i++) {
-				final Component component = ol.get(i);
-				if (!o.contains(component.text())) {
-					continue;
-				}
-				return useKeys ? ctx.input.send(String.valueOf(i + 1)) : component.click("Continue");
+		return null;
```
AnthonyOSX: Looks like you overlooked this (return nil).

ArtificialPB: That's not an issue here because the method is not publicly exposed.

AnthonyOSX: Personally I feel as though even the internals should abide by the same design.

ArtificialPB: This avoids unnecessary wrapper creation.

AnthonyOSX: The nanosecond and two or three allocated bytes on the heap can hardly something justify not following the API paradigm.

ArtificialPB: In my opinion it's clearer that an object does not exist by returning a null value. I understand that the part of the API that's used by third parties needs to null-safe. Making the internals, where it's clear where it's going to be used, null-safe just seems to add unnecessary complexity. 

For example take a look at GenericItem class: https://github.com/powerbot/powerbot/blob/rsbot-api/rt4/GenericItem.java
I don't see any reason why you'd make CacheItemConfig#load() return a null safe wrapper to check if the definition has been loaded or not.

ArtificialPB:
Simplified what @Strikegently noted and ChatOption#select(boolean). Added chat input handling methods.

Timer:
`queryContinue` seems weird to me, why not `canContinue`?

ArtificialPB:
Just made it consistent with rt6 version of this method: https://github.com/powerbot/powerbot/blob/rsbot-api/rt6/Chat.java#L70-L72
I can deprecate rt6 version and add canContinue for both versions, if you'd like?

Timer:
Yeah, deprecating the query and simply making them can would be neater imo.
Keeping this PR open for a while so people can keep discussing.

ArtificialPB:
Deprecated Chat#queryContinue() for rt6 and added canContinue() method. Simplified some rt6 chat stuff.

rt4/Chat.java
Timer: This was written before `ctx.input.send` returned a boolean (it ONLY returns a boolean for these cases [code simplification], pretty sure it always returns true no matter what).

Timer:
Thanks!

PR #1240, opened 2016-01-21
Falador agility v 2.0

achung205:
-double checked script manifest
-consistent variable identifiers
-correct package names
-compiles locally


Strikegently:
Merged.

Issue #1260, opened 2016-01-26
Add support for additional collision flags


ArtificialPB:
Right now, CollisionMap#flagAt(int,int) only returns one of the supported flags (unless I'm missing something), but ignores all other flags on that tile. For example, the tile in the varrock palace, right next to the cage where Monk of Zamorak is in, returns 130 but when loading the flag for the same tile directly from cache returns 545326210.

Timer:
We generate these ourselves and do not extract from the cache. Unfortunately we cannot provide things like range capable etc.
Feel free to make a PR which adds this capability.

ArtificialPB:
Closing this for now, cache region reading does not produce valid collision data for some regions.

Issue #1122, opened 2015-12-22
GameObject.orientation


Overflow-Powerbot:
Access to the orientation for game objects other than wall objects would be useful specifically for generating models/bounds

![Orientation Image](http://puu.sh/m4gvt/aec2043316.png)

NexPB:
This would be awesome :+1:

PR #1262, opened 2016-01-27
Added rt6 region cache reading

ArtificialPB:
https://github.com/powerbot/powerbot/issues/1260

Based off Strikeskids cache reader: https://github.com/Strikeskids/CacheReader

Just putting this up here for discussion. Still need to add additional CollisionFlags.


Logicail:
Aww I was going to take a look at this, I've not had a chance to look through the code yet. But I don't see anything adding objects that are not in the cache (like portable banks)

Does it need a separate method exposing it in Movement, could it replace the existing method (possible with a fallback if cache broken). We may need to check how expensive the call is and cache it (refresh the non cache objects)

It would be a way to add gameobject orientation #1122 (I couldn't find if a local object had an orientation method)

ArtificialPB:
I'll start posting on requests that I'm working on. Please do the same so we wont have multiple people working on the same thing at once.

It currently only loads data from the cache. Should I add support for objects that can be set up by players? The only way I can think of to do that would be by refreshing objects query, which is pretty expensive for the whole region. You have any ideas how to solve this?

I was thinking of replacing the current CollisionMap method in Movement. I'll do that in the end. Reading this map data from cache should be faster than calling the current collisionMap method, but I'm expecting it to be more resource intensive. 

I'll take a look at object orientation.

ArtificialPB:
- Added additional Collision flags.
- Modified ```Movement#collisionMap(int)``` to load collision map from cache. If loading fails it still uses the old method.

Collision flags for objects, that can be set up by players, can be parsed when calling ```CollisionMap#flagAt(int,int)```. Instead of refreshing the whole query, it can be just refreshed on that one tile. For this to work we'd need to have game object orientation hooked.

Anyone got any idea what this flag might be used for: ```0x20000``` Based on Strikeskids cache reader this should be ```OBJECT_BLOCK```.

Timer:
Reading from cache will result in bad paths, as it does not account for doors @ArtificialPB. Please remove the cache usage in movement.

Logicail:
@Timer could it not do a select() and add the doors? Is that not what the collision map method currently does?

ArtificialPB:
Since there is an ```OBJECT_TILE``` flag, why not just check for objects by refreshing the query on those tiles and calculate collision flags @Timer ? If the current method really does select() for the whole region, this might even be faster.

Timer:
Yes the current method `select()`'s the entire region and computes flags (the ones required for walking). If OBJECT_TILE is a reliable way to refresh tiles that require re-evaluation, that'd be neat. But what's their default state and how do we know what to clear?

ArtificialPB:
I did some test and it looks like we'd actually have to load 4 regions to cover the whole loaded map area. Average region load time for me was 22ms so it would take around 80-100ms to load all 4. OBJECT_TILE works just for objects that cover the whole tile so it wont work for doors. For doors we'd have to use BLOCK_"DIRECTION" and check those tiles, and if possible, use current collision flag calculations to get the actual flag.

Strikeskids:
A solution I toyed with previously was not setting the collision flags for transient objects (doors, gates) from the cache at all. It's easy enough to determine where the transient objects could be located. You could just make up another flag for possibly transient and query on those.

ArtificialPB:
How would you determine that @Strikeskids ? The only way I can currently think of to do that is like I said in previous post, by re-checking tiles with BLOCK_"DIRECTION" flags. I did however notice that some door, gate objects have ALLOW_RANGE flags, but it's not reliable.

Strikeskids:
I guess the way I was doing it was via object name and actions filter. Because I was doing massive pre-processing on the entire map, that worked fine. I don't know if that's feasible for this scenario though.

Timer:
I don't know if we can account for every situation using names ... I think this would be better left working as is.
If anything @ArtificialPB, we can strip out ALL flags that pertain to walkability or direction then just and the ones we compute with the cache so there's properties like allow range, etc. I mean what are we really gaining from this? The ability to check if you can use a bow over an area? What's the significance of the BLOCK_NORTH (etc) vs just normal NORTH, I don't recall and need a refresher.

ArtificialPB:
I agree with you @Timer about finding these objects by using names. Well it would make my code (and possibly others) a bit cleaner and it would save me some time because I wouldn't have to write additional classes to account for example for npcs in cages and things like that. It would also make accessing collision map data faster compared to the current method. This is not a high priority addition though (unlike https://github.com/powerbot/powerbot/pull/1268 ;) ). I suggest leaving this pull request open for now to see if we come up with a better solution.

I think the main difference between those two flags is that one allows you to stand on that tile (e.i. NORTH), and the other one blocks it altogether (BLOCK_NORTH).

ArtificialPB:
@Logicail Actually you can walk on tiles where portables are set, at least for portable forges. I did not test any other yet.

Polish-Civil:
`package me.polishcivil.powerbot.api.region

import me.polishcivil.powerbot.api.utils.ObjectType.*
import me.polishcivil.powerbot.api.utils.extensions.isDoor
import me.polishcivil.powerbot.api.utils.extensions.toLocal
import me.polishcivil.powerbot.api.utils.extensions.type
import org.powerbot.script.rt6.ClientContext
import org.powerbot.script.rt6.GameObject

/**
 * Created by polish on 19.02.16.
 */
object LocalRegion {
    private var cachedSettings: Array<TileSettings>? = null
    private var cachedSettingsBaseX = -1
    private var cachedSettingsBaseY = -1
    private var cachedSettingsInitTime = -1L

    fun calculateTileSettings(ctx: ClientContext): Array<TileSettings> {
        val mapOffset = ctx.client().world.mapOffset
        if (mapOffset.x == LocalRegion.cachedSettingsBaseX && LocalRegion.cachedSettingsBaseY == mapOffset.y) {
            if (System.currentTimeMillis() - cachedSettingsInitTime <= 600) {
                return LocalRegion.cachedSettings!!
            }
        }

        val floorSettings = ctx.client().world.floorSettings.bytes
        val tiles = ctx.client().world.landscape.tiles
        val floors = floorSettings.size

        val width = 104
        val height = 104
        val settings = Array(floors, { TileSettings(width, height) })


        if (floorSettings.isNotEmpty() && tiles.isNotEmpty()) {
            for (floor in 0..floors - 1) {
                for (x in 0..width - 1) {
                    for (y in 0..height - 1) {
                        val flag = floorSettings[floor][x][y].toInt()
                        if (flag and 0x1 != 0) {
                            val bridgeFlag = floorSettings[1][x][y].toInt()
                            val bridgeFloor = if (bridgeFlag and 0x2 == 0) floor else floor - 1
                            if (bridgeFloor >= 0) {
                                settings[bridgeFloor].setWaterMask(x, y)
                            }
                        }
                    }
                }
            }
            val objects = Array(width) { Array(height, { Array(floors, { mutableListOf<GameObject>() }) }) }
            ctx.objects.select().forEach {
                val local = it.tile().toLocal(ctx)
                if (local.x() >= 0 && local.y() >= 0)
                    if (local.x() < width && local.y() < height) {
                        if (!it.isDoor()) {
                            objects[local.x()][local.y()][local.floor()].add(it)
                        }
                    }
            }
            for (floor in 0..floors - 1) {
                LocalRegion.visitClipMapObjects(settings[floor], floor, floorSettings, settings, objects)
            }
        }

        cachedSettings = settings
        cachedSettingsBaseX = mapOffset.x
        cachedSettingsBaseY = mapOffset.y
        cachedSettingsInitTime = System.currentTimeMillis()
        return settings
    }

    private fun visitClipMapObjects(startTileSettings: TileSettings, startFloor: Int, floorSettings: Array<Array<ByteArray>>, tileSettingsArray: Array<TileSettings>, objects: Array<Array<Array<MutableList<GameObject>>>>) {

        var floor = startFloor
        for (x in 0..startTileSettings.sizeX - 1) {
            for (y in 0..startTileSettings.sizeY - 1) {
                if (floorSettings[startFloor][x][y].toInt() and 0x2 == 0x2) {
                    floor = 1
                }
                val gameObjects = objects[x][y][floor]
                gameObjects.forEach {
                    tileSettingsArray[floor].unsetMask(x, y, LocalCollisionData.UNDEFINED)
                    if (it.`object`.type != -1) {
                        LocalRegion.visitClipMapObject(tileSettingsArray[floor], x, y, it)
                    }
                }
            }
        }
    }

    private fun visitClipMapObject(tileSetting: TileSettings, x: Int, y: Int, gameObject: GameObject) {
        //TODO
        val basicObject = gameObject.`object`
        val clip = gameObject.clippingType()
        val width = 1
        val height = 1
        val blocked = false
        val type = basicObject.type()
        val orientation = gameObject.orientation()
        if (type == GROUND_DECORATION) {
            if (clip == 1) {
                tileSetting.setMask(x, y, 0x40000);
            }
        } else {
            if (clip != 0) {
                if (type == GENERAL_OBJECT || type == GROUND_OBJECT) {
                    tileSetting.setObjectMasks(blocked, width, height, x, y, orientation)
                } else if (type.ordinal >= 12) {
                    tileSetting.setObjectMasks(blocked, width, height, x, y, orientation)
                } else if (type.ordinal >= 0 && type.ordinal <= 3) {
                    //walls
                    tileSetting.setWallMasks(x, y, orientation, type.ordinal, blocked)
                } else if (type == WALL_DIAGONAL) {
                    tileSetting.setObjectMasks(blocked, width, height, x, y, orientation)
                }
            }
        }
    }
}`

`package me.polishcivil.powerbot.api.region

object LocalCollisionData {
    val BASE_LENGTH = 104
    val BASE_BOUNDARY = LocalCollisionData.BASE_LENGTH - 1

    val WALL_NORTH_WEST = 1
    val WALL_NORTH = 2
    val WALL_NORTH_EAST = 0x4
    val WALL_EAST = 0x8
    val WALL_SOUTH_EAST = 0x10
    val WALL_SOUTH = 0x20
    val WALL_SOUTH_WEST = 0x40
    val WALL_WEST = 0x80
    val BLOCKED = 0x100
    val WATER = 0x200000
    val UNDEFINED = 0x1000000
}
`
`package me.polishcivil.powerbot.api.utils.extensions

import org.powerbot.script.rt6.GameObject

/**
 * Created by polish on 19.02.16.
 */
fun GameObject.isDoor(): Boolean = hasAction("open") || name().equals("door", ignoreCase = true) || name().equals("gate", ignoreCase = true)

fun GameObject.hasAction(action: String): Boolean {
    val actions = actions()
    for (s in actions) {
        if (s != null && s.equals(action, ignoreCase = true)) {
            return true
        }
    }
    return false
}`
`package me.polishcivil.powerbot.api.utils.extensions

import me.polishcivil.powerbot.api.utils.ObjectType
import org.powerbot.script.rt6.BasicObject

/**
 * Created by polish on 19.02.16.
 */
fun BasicObject.type(): ObjectType = ObjectType.values()[type]
`
`package me.polishcivil.powerbot.api.utils.extensions

import org.powerbot.script.Tile
import org.powerbot.script.rt6.ClientContext

/**
 * Created by polish on 19.02.16.
 */

fun Tile.toLocal(ctx: ClientContext): Tile = Tile(tile().x() - ctx.client().world.mapOffset.x, tile().y() - ctx.client().world.mapOffset.y, tile().floor())
fun Tile.toGlobal(ctx: ClientContext): Tile = Tile(tile().x() + ctx.client().world.mapOffset.x, tile().y() + ctx.client().world.mapOffset.y, tile().floor())
`
`package me.polishcivil.powerbot.api.region

class TileSettings
/**
 *
 */
(val sizeX: Int, val sizeY: Int) {
    val clipData: Array<IntArray>

    init {
        clipData = Array(this.sizeX) { IntArray(this.sizeY) }
        init()
    }

    /**

     */
    fun init() {
        for (i in 0..sizeX - 1) {
            for (j in 0..sizeY - 1)
                setMask(i, j, LocalCollisionData.UNDEFINED)
        }

    }

    /**
     * @param x
     * *
     * @param y
     * *
     * @param orientation
     * *
     * @param objectType
     * *
     * @param flag
     */
    fun setWallMasks(x: Int, y: Int, orientation: Int, objectType: Int, flag: Boolean) {

        if (objectType == 0) {
            if (orientation == 0) {
                setMask(x, y, 128)
                setMask(x - 1, y, 8)
            }
            if (orientation == 1) {
                setMask(x, y, 2)
                setMask(x, y + 1, 32)
            }
            if (orientation == 2) {
                setMask(x, y, 8)
                setMask(x + 1, y, 128)
            }
            if (orientation == 3) {
                setMask(x, y, 32)
                setMask(x, y - 1, 2)
            }
        }
        if (objectType == 1 || objectType == 3) {
            if (orientation == 0) {
                setMask(x, y, 1)
                setMask(x - 1, y + 1, 16)
            }
            if (orientation == 1) {
                setMask(x, y, 4)
                setMask(x + 1, y + 1, 64)
            }
            if (orientation == 2) {
                setMask(x, y, 16)
                setMask(x + 1, y - 1, 1)
            }
            if (orientation == 3) {
                setMask(x, y, 64)
                setMask(x - 1, y - 1, 4)
            }
        }
        if (objectType == 2) {
            if (orientation == 0) {
                setMask(x, y, 130)
                setMask(x - 1, y, 8)
                setMask(x, y + 1, 32)
            }
            if (orientation == 1) {
                setMask(x, y, 10)
                setMask(x, y + 1, 32)
                setMask(x + 1, y, 128)
            }
            if (orientation == 2) {
                setMask(x, y, 40)
                setMask(x + 1, y, 128)
                setMask(x, y - 1, 2)
            }
            if (orientation == 3) {
                setMask(x, y, 160)
                setMask(x, y - 1, 2)
                setMask(x - 1, y, 8)
            }
        }
        if (flag) {
            if (objectType == 0) {
                if (orientation == 0) {
                    setMask(x, y, 0x10000)
                    setMask(x - 1, y, 4096)
                }
                if (orientation == 1) {
                    setMask(x, y, 1024)
                    setMask(x, y + 1, 16384)
                }
                if (orientation == 2) {
                    setMask(x, y, 4096)
                    setMask(x + 1, y, 0x10000)
                }
                if (orientation == 3) {
                    setMask(x, y, 16384)
                    setMask(x, y - 1, 1024)
                }
            }
            if (objectType == 1 || objectType == 3) {
                if (orientation == 0) {
                    setMask(x, y, 512)
                    setMask(x - 1, y + 1, 8192)
                }
                if (orientation == 1) {
                    setMask(x, y, 2048)
                    setMask(x + 1, y + 1, 32768)
                }
                if (orientation == 2) {
                    setMask(x, y, 8192)
                    setMask(x + 1, y - 1, 512)
                }
                if (orientation == 3) {
                    setMask(x, y, 32768)
                    setMask(x - 1, y - 1, 2048)
                }
            }
            if (objectType == 2) {
                if (orientation == 0) {
                    setMask(x, y, 0x10400)
                    setMask(x - 1, y, 4096)
                    setMask(x, y + 1, 16384)
                }
                if (orientation == 1) {
                    setMask(x, y, 5120)
                    setMask(x, y + 1, 16384)
                    setMask(x + 1, y, 0x10000)
                }
                if (orientation == 2) {
                    setMask(x, y, 0x5000)
                    setMask(x + 1, y, 0x10000)
                    setMask(x, y - 1, 1024)
                }
                if (orientation == 3) {
                    setMask(x, y, 0x14000)
                    setMask(x, y - 1, 1024)
                    setMask(x - 1, y, 4096)
                }
            }
        }
    }

    /**
     * @param flag
     * *
     * @param objectSizeX
     * *
     * @param objectSizeY
     * *
     * @param objectX
     * *
     * @param objectY
     * *
     * @param orientation
     */
    fun setObjectMasks(flag: Boolean, objectSizeX: Int, objectSizeY: Int, objectX: Int, objectY: Int, orientation: Int) {
        var objectSizeX = objectSizeX
        var objectSizeY = objectSizeY
        var mask = 256
        if (flag)
            mask += 0x20000
        if (orientation == 1 || orientation == 3) {
            val l1 = objectSizeX
            objectSizeX = objectSizeY
            objectSizeY = l1
        }
        for (x in objectX..objectX + objectSizeX - 1)
            if (x >= 0 && x < sizeX) {
                for (y in objectY..objectY + objectSizeY - 1)
                    if (y >= 0 && y < sizeY)
                        setMask(x, y, mask)

            }

    }

    /**
     * @param x
     * *
     * @param y
     * *
     * @param mask
     */

    fun unsetMask(x: Int, y: Int, mask: Int) {
        if (x < sizeX && y < sizeY && x >= 0 && y >= 0) {
            this.clipData[x][y] = this.clipData[x][y] and mask.inv()
        }
    }

    /**
     * @param x
     * *
     * @param y
     * *
     * @param mask
     */
    fun setMask(x: Int, y: Int, mask: Int) {
        if (x < sizeX && y < sizeY && x >= 0 && y >= 0) {
            clipData[x][y] = clipData[x][y] or mask
        }
    }

    /**
     * @param x
     * *
     * @param y
     */
    fun setWaterMask(x: Int, y: Int) {
        if (x < sizeX && y < sizeY && x >= 0 && y >= 0) {
            setMask(x, y, LocalCollisionData.WATER)
        }
    }
}`

Polish-Civil:
My approach:

[region.zip](https://github.com/powerbot/powerbot/files/137520/region.zip)
[extensions.zip](https://github.com/powerbot/powerbot/files/137521/extensions.zip)

I used what it's in api, so few things are missing but it works

ArtificialPB:
@Polish-Civil It looks like what @Strikeskids suggested, to find doors and gates using names/actions, unless I'm missing something? And this won't work for dynamic objects, unless we load their definitions too.

Polish-Civil:
@ArtificialPB Dynamic objects?

ArtificialPB:
Objects that have different states. One example of those are Lodestones, which must be activated before you can use them. @Polish-Civil

Polish-Civil:
@ArtificialPB Ok, cool, but how to get them then if ctx.objects doesn't grab em.

ArtificialPB:
Totally forgot that Rsbot now loads child definitions instead of parent dynamic object, so scratch that last part about loading child definitions in my first reply @Polish-Civil You can get everything now using objects query.

Polish-Civil:
@ArtificialPB Oh, ok well then. So object definition additions and my thingy gonna be working fully.
But I see powerbot doesn't give a f. about anything

Timer:
@Polish-Civil why you coming in here and being so rude?

Polish-Civil:
@Timer Because im wasting my time waiting for such easy and obvious requests. No offence obv,, but i can be mad.

Timer:
@ArtificialPB I'm going to take a peek at this soon and see what we can do with integrating it.

ArtificialPB:
:+1: @Timer 

@Logicail @Strikeskids do you maybe know what flag ```0x200000``` is used for?

Polish-Civil:
@ArtificialPB Water

ArtificialPB:
Are you sure it's just water @Polish-Civil ? This flag is also returned for unreachable areas like those on top of the hills that surround Al'Kharid.

Polish-Civil:
@ArtificialPB I use such naming from osrs, such flag is settled there only for bridges, so i call it water

Strikeskids:
It looks like that flag occurs if there is an object pull-down (when objects from higher planes get pulled down to lower planes).

https://github.com/Strikeskids/CacheReader/blob/b0206fc02475b8884fcfbfb06bd47aace8e85e68/src/com/sk/cache/wrappers/region/Region.java#L46

ArtificialPB:
Closing this for now, cache region reading does not produce valid collision data for some regions.

PR #1278, opened 2016-02-02
Blast furnace v 1.3: better movement, randomized cam movement, handles dialogue from misclicks

achung205:
v 1.3
now moves to bank if distance > 5 (instead of just in viewport)
now moves to belt if distance >= 5 (instead of just in viewport)
randomized camera movement (turnTo)
removed random npc handling
handles dialogue


Strikegently:
You're importing obfuscated classes (z.RA). You need to remove that; otherwise your script will break in the next client version.

achung205:
fixed

PR #1282, opened 2016-02-05
Blast Furnace v1.3c : fixed logic with moving to conveyor belt that caused script to stop

achung205:
* v 1.3c
 * fixed some logic while moving to conveyor belt
 * randomized camera turning decision
 * handles case when attempting to deposit ores when furnace is full
 * screenshot on script end


ChrisBPB:
merged

PR #1286, opened 2016-02-05
Removed old thieving script and added a new OSRS script

helicopter12:
Deleted the old EOC Runescape script and added a re-visted version for OSRS which is still underdevelopment since I am without membership. The script is hidden so that Chris can assign users access to the script for development and testing.


ChrisBPB:
merged

Strikegently:
@helicopter12 
Your script doesn't compile. This is why we urge people to test their scripts before pushing the changes.

PR #1307, opened 2016-02-17
Blast Furnace v 1.4c: fixed foreman payment when 60+

achung205:
v 1.4c
fixed foreman payment when 60+
fixed a bug while collecting that should be handled by idle timer
added a check for coins subtracted to confirm payment
now checks exact amount of bars collected (assumed full load before)
urgent.


ChrisBPB:
merged

PR #1311, opened 2016-02-19
Blast furnace v 1.4d: bug fix and code tidying

achung205:
v 1.4d
fixed a bug that repeatedly opens bank when attempting withdraw coins
removed useless setBarType function
code formatting

### DIFF
--- a/src/andyroo/blastfurnace/BlastFurnace.java
+++ b/src/andyroo/blastfurnace/BlastFurnace.java
@@ -522,6 +522,10 @@ public class BlastFurnace extends PollingScript<ClientContext> implements PaintL
      *
      * @return number of ores withdrawn, -1 if failed
      */
+
+    /*
+    IMPROVE THIS
+     */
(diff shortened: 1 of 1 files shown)


src/andyroo/blastfurnace/BlastFurnace.java
```diff
@@ -811,11 +807,12 @@ public Boolean call() throws Exception {
...
             }
-        }
-        else {
+        } else {
             log.info("Going to withdraw coins");
-            if(openBank()) {
-                if(ctx.inventory.select().count() == 28) {
+            if (openBank()) {
+                System.out.println(ctx.inventory.select().count());
+                if (ctx.inventory.select().count() == 28) {
```
Timer: Since you selected above for the sout, this can be reduced to just `if (ctx.inventory.count() == 28) {`

achung205:
i see what youre saying.

PR #1313, opened 2016-02-20
Blast Furnace v 1.4e: fixed screenshot

achung205:
v 1.4e
fixed screenshot having a black background, hiding the black text
fixed bug where script attempted to click bank chest while bar widget was open


ChrisBPB:
As a script writer you should really upload this to your private repository now because you need to upload stuff there to avoid automatic demotion for inactivity.

achung205:
haha im having trouble setting it up...

Issue #847, opened 2015-03-30
Implement grandexchange notification ?


swatarianess:
The Grandexchange Hook has been the same for more than half a decade. Would it be possible to implement it? (RT4 and RT6)

Request: Implement the hook.

It contains information such as what items are being sold, amount sold, percentage complete etc. Would be good to see it utilized by scripts situated around Varrock.

benpoulson:
We have upcoming work to implement a big backlog of new and missing hooks.
It's a contiversial area to hook, but if @Paris is happy for us to do so we can add GE hooks into the TODO list. @AshGreh 

For now I'm closing this ticket. However, I've made a note to comment on it when these specific hooks has been implemented.

PR #1318, opened 2016-02-22
Implement Grand Exchange

Moosheplusplus:
Added support for the Grand Exchange per request of @AnthonyOSX.


Timer:
Pinging some other developers, how do you guys feel about the creation of an entirely new event for this?
Should it be done in this fashion?

@AnthonyOSX @ArtificialPB @Strikegently @Logicail

AnthonyOSX:
I think an entirely new event for what appears to be a concise MessageListener seems a bit overkill. A new event ID for a Grand Exchange event could be added so users could just check the event ID and handle the message as they please.

rt4/GrandExchange.java
```diff
@@ -0,0 +1,316 @@
...
+import org.powerbot.script.Condition;
+import org.powerbot.script.MessageEvent;
+import org.powerbot.script.MessageListener;
+import org.powerbot.script.Random;
+import org.powerbot.script.rt4.*;
+import org.powerbot.script.rt4.Game.Crosshair;
+
+public class GrandExchange extends ClientAccessor implements MessageListener {
+	
+	public static final int[] GE_CLERK = {2148, 2149, 2150, 2151};
```
AnthonyOSX: Determining exchange NPC's should probably be done by using actions instead of static ID's. Future proof :)

ArtificialPB: Finding them by name would be better. There are other npcs in game with "Exchange" action (Tool Leprechauns, Void Knights, etc...). Npc and object ids change after every update in OSRS.

rt6/GrandExchange.java
```diff
@@ -0,0 +1,332 @@
...
+import org.powerbot.script.MessageEvent;
+import org.powerbot.script.MessageListener;
+import org.powerbot.script.Random;
+import org.powerbot.script.rt6.*;
+import org.powerbot.script.rt6.Item;
+import org.powerbot.script.rt6.Game.Crosshair;
+
+public class GrandExchange extends ClientAccessor implements MessageListener {
+	
+	public static final int[] GE_CLERK = {1419, 2593, 2240, 2241};
```
ArtificialPB: You're missing some ids: 15529, 15530, 19911, 19912, 19913, 19914, 19915

ArtificialPB:
If https://github.com/powerbot/powerbot/issues/847 gets implemented you could completely ditch the GeEvent/MessageEvent implementation and provide a query for offers.

rt4/GrandExchange.java
AnthonyOSX: An action check would be much more generic.

Strikegently:
Unless we get the GE hooks for more complete information, this is fine.

Timer:
I personally don't like the event-based notifications... maybe we should look into doing #847 if developers feel they really want a Grand Exchange API.

Moosheplusplus:
@Timer it would be a better solution than using a MessageListener to dispatch events

rs-sliske:
ge hooks would be better than message listener

the buy methods need to find the item by id, not name, there are too many name collisions, and the ge sorts alphabetically, not by best match (even if you type the name exactly it wont always be the first item in the list)

im working on version that uses id atm.

Moosheplusplus:
Is everyone okay with just removing GeEvent and implement the hook at another time, so people can still use the interface for the grand exchange? That way this can be merged and be ready for v7025?

AnthonyOSX:
@Moosheplusplus yeah remove GE event. Anyone that needs it for now can just use the message handler.

ArtificialPB:
Fine with me @Moosheplusplus

swatarianess:
I have the Grand-Exchange hook if anyone wants it. I have had it for almost a month.

AnthonyOSX:
@swatarianess your hook is only applicable for the current game revision. One of the main developers needs to hook it within the updater.

Moosheplusplus:
@TImer @Paris This Pull Request is ready for v7026 and does not require any hooks. This will serve as an interface for the Grand Exchange, until the proper hooks have been implemented.

Strikegently:
If we're going to do it, we should do it right the first time rather than coming back to clean up.
We can just postpone a Grand Exchange API until we get the hooks.

@Timer

PR #1323, opened 2016-02-23
Made minimap components future proof

ArtificialPB:
Title


Moosheplusplus:
I like the idea on not depending on the component index, however, for the other components in the minimap, such as run/energy and teleportation, they have no special types, so we can't easily get them that way. The compass, though, has a special type of 1339.

What do you think about relying on texture ID?

ArtificialPB:
Yeah we can find other components by texture id. I'd just make minimap and compass parsing using special types, because it's seems to be the most accurate. I'll update this pull request.

Timer:
I was going to do this before but for some reason 1338 didn't always work ... have they changed this?

ArtificialPB:
I've checked and we can find minimap by special type in both legacy and rs3 interface modes. I've als oupdated my previous post @Timer

ArtificialPB:
Now finds minimap comp by special type, and all other blocking components by their bounds. Bounds do not change if you resize the minimap.

Timer:
I had to remove the `component.contentType() != 1338` check at one point because it had broken the minimap. We can give this another try.

ArtificialPB:
@Timer I can add ```ctx.widgets.component(Constants.MOVEMENT_WIDGET, Constants.MOVEMENT_MAP)``` if parsing by contentType() fails, if you'd like?

Timer:
Feel free, @ArtificialPB

PR #1325, opened 2016-02-23
Exposed Lobby#World component index

ArtificialPB:
This is useful for in-game world hopping. Feel free to rename ```componentIndex()``` to ```index()```. I've named it like that so people don't confuse it with ```number()``` method.


Timer:
Merged in 184bdb2

PR #1327, opened 2016-02-24
Autonomous

AutonomousCoding:
Modified antiban; Added logout on script stop; Moved time from state to repaint for more accurate time;


ChrisBPB:
merged

PR #1336, opened 2016-03-02
Support for different textures for game tabs

ArtificialPB:
Quest tab can have different textures depending on what the user selected to show (minigames, achievement diary, etc...). This is rt4.


ArtificialPB:
Added support for logout tab when bottom line tabs interface setting is selected.

PR #1343, opened 2016-03-06
Delete from public repo

achung205:
Delete humidify script from public repo and see if it appears in private repo


Strikegently:
Remove DistanceLogger. Your test scripts shouldn't be there.

Strikegently:
To be clear, you know this deletes all of your scripts, correct?

achung205:
i want to delete the scripts from the public repo. ive already uploaded them to my developer repo, so there are two versions on the sdn right now.

PR #1349, opened 2016-03-11
Rob's Tutorial Island: added in client details

Montezumas:
added in client details so that the script appears under OS in the SDN


Strikegently:
In the future, include the script name and commit purpose in your pull request titles.

PR #1351, opened 2016-03-13
Added Html remover for object and npc names

ArtificialPB:
Since Menu actions/options and items are stripped of html, it seems reasonable to remove them from objects and npcs too to avoid issues when interacting.


Moosheplusplus:
do any Objects/NPCs actually have any tags in their names?

ArtificialPB:
Yes they do, that's why I've made this PR. An example of this would be the Portal NPC to exit rune essence mine.

PR #1374, opened 2016-04-12
Added Grand Exchange Alcher

ChrisHemmens:
Added Grand Exchange Alcher: https://www.powerbot.org/community/topic/1307713-grand-exchange-alcher/


Strikegently:
1. the "topic" flag in your manifest should just be the topic ID, not the URL.
2. Is it possible to make the script without jsoup? Large third-party dependencies are ill-advised.

ChrisHemmens:
@Strikegently No jsoup is required. I updated the manifest.

PR #1378, opened 2016-04-15
Add support for placeholders

Moosheplusplus:


ChrisBPB:
Someone said there was another update where they moved the lock button but I just checked and the constants updated in this file are the correct. :)

PR #1380, opened 2016-04-16
Added mistermaru/CanifisAgilityCourse

Mistermaru:
Now uses enums instead of JumpGap1-6


Strikegently:
Your script needs a topic.

PR #1399, opened 2016-04-23
Added Production Interface utilities

ArtificialPB:
Added utilities for interacting with Production Interface: http://runescape.wikia.com/wiki/Make-X


Strikegently:
This seems like something that script writers should handle, not necessarily the API.

ArtificialPB:
I think this will be an useful addition to the API and it will make script writing easier, especially for new users. This is like the Grand Exchange API. Why make everyone write their own API or copy it from snippets when it can be included in the RSBot API and be available from the start?

Moosheplusplus:
Almost every preparation skill uses this interface, so I think it would be a valid addition for the RSBot API.

Timer:
@Strikegently do you stand against not allowing this in or would it be a valid addition?

ArtificialPB:
Solved conflicts with the base branch

PR #1462, opened 2016-06-27
Improve Magic Interface

Moosheplusplus:
- Added validity detection for Spells (which compares the current component's texture ID with the spell's natural "off" texture ID)
- Fixed Bones to Peaches and Earth Strike
- Added documentation
- Declared explicit access modifiers for constructors
- Added Ancient Magicks
- Future-proof component indexes

**Suggestions**
- ~~Perhaps add the "on" texture IDs so we don't have to rely on component IDs?~~ Done.

### DIFF
--- a/src/org/powerbot/script/rt4/Magic.java
+++ b/src/org/powerbot/script/rt4/Magic.java
@@ -28,10 +28,10 @@ public class Magic extends ClientAccessor {
 	public Spell spell() {
 		final Book book = book();
 		for (final Spell spell : Spell.values()) {
-			if (spell.book != book) {
+			if (spell.book() != book) {
 				continue;
 			}
-			if (ctx.widgets.component(spell.book.widget, spell.component()).borderThickness() == 2) {
+			if (ctx.widgets.component(spell.book().widget, spell.component()).borderThickness() == 2) {
 				return spell;
 			}
 		}
@@ -70,14 +70,12 @@ public class Magic extends ClientAccessor {
 	 * @param spell The spell to find the component for.
 	 * @return The resulting component
 	 */
-	public Component component(final MagicSpell spell) {
+	public Component component(final MagicSpell s) {
 		Widget w = ctx.widgets.widget(Book.widget());
 		for(Component c : w.components()) {
 			int texture = c.textureId();
-			for(MagicSpell s : book().spells) {
-				if(texture == s.texture() || texture == s.texture() - 50)
-					return c;
-			}
+			if(texture == s.texture() || texture == s.texture() - 50)
+				return c;
 		}
 		return w.component(-1);
 	}
@@ -107,7 +105,7 @@ public class Magic extends ClientAccessor {
 	 * Modern (Standard) Spells
 	 */
 	public enum Spell implements MagicSpell {
-		NIL(-1, -1),//Selected spell 192,x=bt2
+		NIL(Integer.MIN_VALUE, Integer.MIN_VALUE),//Selected spell 192,x=bt2
 		HOME_TELEPORT(0, 406),
 		WIND_STRIKE(1, 1, 65),
 		CONFUSE(3, 2, 66),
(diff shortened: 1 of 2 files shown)


ArtificialPB:
Castable spell texture is always uncastable texture - 50 (in your case, onTexture = offTexture - 50). You can ditch component indexes and use only textures.

src/org/powerbot/script/rt4/Magic.java
```diff
@@ -3,22 +3,28 @@
...
+	/**
+	 * Retrieves the current book.
+	 * 
+	 * @return The current book.
+	 */
 	public Book book() {
-		for (final Book b : Book.values()) {
-			if (ctx.widgets.widget(b.widget).component(0).valid()) {
+		for(Book b : Book.values()) {
+			if((ctx.varpbits.varpbit(Constants.SPELLBOOK_VARPBIT) & 0x3) ==
```
ArtificialPB: Introduce a local variable for varp value so you don't have to grab it each iteration.

src/org/powerbot/script/rt4/Magic.java
```diff
@@ -47,120 +63,265 @@ public boolean poll() {
...
+	/**
+	 * Grabs the component for the given MagicSpell.
+	 * 
+	 * @param spell The spell to find the component for.
+	 * @return The resulting component
+	 */
+	public Component component(MagicSpell spell) {
+		Widget w = ctx.widgets.widget(Book.widget());
+		if(spell instanceof Spell)
+			return w.component(((Spell) spell).component);
```
ArtificialPB: You can remove lines 102 and 103 because you already find components by texture.

src/org/powerbot/script/rt4/Magic.java
```diff
@@ -47,120 +63,265 @@ public boolean poll() {
...
+		
+		private Spell(final int level, final int component,
+				final int offTexture) {
 			this.level = level;
 			this.component = component;
+			this.offTexture = offTexture;
 		}
 
+		@Deprecated
 		public Book book() {
```
ArtificialPB: Why is this deprecated? You should add this method to MagicSpell interface so users can check to which spell book certain spells belong.

src/org/powerbot/script/rt4/Magic.java
```diff
@@ -47,120 +63,265 @@ public boolean poll() {
...
+	}
+	
+	/**
+	 * Determines whether or not the specified spell is ready to be casted.
+	 * 
+	 * @param spell The spell to validate.
+	 * @return {@code true} if it is ready to be cast, {@code false} otherwise.
+	 */
+	public boolean ready(final Spell spell) {
+		return component(spell).textureId() != spell.texture();
```
ArtificialPB: As far as I know, textures only update if the tab is opened.

ArtificialPB:
Also a lot of local variables and method parameters can be made final.

src/org/powerbot/script/rt4/Magic.java
Moosheplusplus: You will not be casting a spell outside of the Magic Tab, so it's not necessary to update. Spells such as Alchemy and Bones to Bananas switch back to the Magic Tab and update after it is cast, so it is not an issue.

src/org/powerbot/script/rt4/Magic.java
```diff
@@ -107,7 +105,7 @@ public Component component(final MagicSpell spell) {
 	 * Modern (Standard) Spells
 	 */
 	public enum Spell implements MagicSpell {
-		NIL(-1, -1),//Selected spell 192,x=bt2
+		NIL(Integer.MIN_VALUE, Integer.MIN_VALUE),//Selected spell 192,x=bt2
```
ArtificialPB: Don't use those values. On line 77, where you check for texture and decrement it by 50, you'll get a positive number instead due to numeric overflow. Change it back to -1 and check if the component texture is not -1 on line 77. And like I said before, a lot of local variables can be made final (line 76 for example).

ArtificialPB:
Add a method to get the currently selected spell.

And I still think you should deprecate the whole "Spell" enum and make a new one for Standard magic spells. NIL spell should be an anonymous class instance of MagicSpell interface.

Moosheplusplus:
@ArtificialPB thanks for the component textures. Do you possibly have the magic values for the Arceuus Spellbook?

Timer:
So what's going on with this?

Moosheplusplus:
@Timer everything about it is fine. It can be merged.

Timer:
@Moosheplusplus can you please rebase this?

Moosheplusplus:
@Timer no

Timer:
yeah I see you did that completely wrong @Moosheplusplus

PR #1468, opened 2016-07-07
Support Genie and Drunken Dwarf in randoms

Moosheplusplus:
Rather than dismissing Genie and Drunken Dwarf, player will interact with the NPCs to collect their items.


rs-sliske:
agree with the genie, although it would also mean bringing back the option for what skill to put lamps in

dwarf only gives junk

rick turpentine normally gives coins / gems which could be useful

Timer:
I'm not a fan of this, let's leave this up to scripts.

Moosheplusplus:
Scripts don't have any control over it, though

PR #1473, opened 2016-07-11
Add multi-combat boolean (rt4)

Moosheplusplus:


ArtificialPB:
Shift and mask the value you get. Only one bit is used to determine if you're in multi-way combat and there are most likely other uses of this varpbit so it can return a different value even though multi-way combat is active.

Here's a good web binary converter to help you with this: http://www.mathsisfun.com/binary-decimal-hexadecimal-converter.html

And in my opinion, the Combat, rather than the Game class, is better suited for this method placement.

Moosheplusplus:
cba.

Issue #1476, opened 2016-07-17
Fix Item interaction in rt4


Moosheplusplus:
![](https://i.imgur.com/VPhF6co.png)

`Item#nextPoint()` is red. According to @Strikegently, it's not supposed to be randomly generated like this.

Timer:
Closing for #1477

PR #1477, opened 2016-07-17
Use gaussian distribution for Item#nextPoint()

Moosheplusplus:
Addresses #1476


Timer:
This must use the inbuilt calculations class https://github.com/powerbot/powerbot/blob/rsbot-api/src/org/powerbot/script/rt4/Component.java#L305-L313

Timer:
Implemented per dd47b576148520b795f762659e4b124a6ff94018

PR #1488, opened 2016-08-14
new CHAT_CONTINUES widget

NoWastePB:
I came across an chat widget with click to continue but it wasn't supported, added it.


Strikegently:
Make the spacing consistent with the rest of the class.

PR #1501, opened 2016-09-07
Make rt4 Actor#healthPercent() public

ArtificialPB:
Since other methods for getting Actors health are deprecated and now use healthPercent() method, it should be made public.


Timer:
Nice catch! This was meant to be public.

PR #1502, opened 2016-09-15
Future-proofing rt4 Combat#health()

ArtificialPB:
No need to rely on components, current Hitpoints level decreases when you get hit.


Timer:
This will be covered in #1527.

Issue #1373, opened 2016-04-10
Disable dismissing Random Events if the script is paused or input is allowed.


Moosheplusplus:
If input is allowed or the script is paused, that's typically an indication of the player themselves controlling the client, therefore, if a random event appears, the player will fight over the input with the bot.

Paris:
In light of #1386 do you think this is still required?

Strikegently:
@Paris 
Yes, but only if mouse control is allowed. That would probably be best, I think.

Moosheplusplus:
Fixed in #1512.

Issue #1362, opened 2016-04-04
Disable dismissing random events via Script Manifest properties


Moosheplusplus:
I request a property to disable the bot from dismissing random events automatically, so the script can handle certain random events itself.

Paris:
Which random event specifically because there was a property for disabling widget closer.

Strikegently:
@Paris 
He asked for "certain" events, so I assume he wants a property for each random event individually.
We could probably have a property for each. It might get messy, though.

Potentially clean it up by adding a constants for all the available properties.

Moosheplusplus:
for example, the ones that could be solved with just a single conversation (such as the genie or mysterious man).

Paris:
In that case would it not be better for the bot to handle those conversations itself?

PR #1512, opened 2016-10-04
Ability to disable the random event handler

stumpy3toes:
By setting the client property "randomevents.disable" to "true" a script writer can implement their own handling for random events. It's simple and optional and resolves this #1362. Also if the user has control of the mouse the random event handler is disabled #1373


Timer:
Why don't we just fix #1593?
Can we compile a list of cases to not run the handler under?

Timer:
What is the significance of ctx.input.blocking()?

Moosheplusplus:
no.

This has many practical uses; if a developer wants to collect genie bottles and use the XP, it can. This just happens to also fix #1593.

PR #1513, opened 2016-10-06
Updated Grand Exchange Alcher

ChrisHemmens:
Updated paint
fixed bugs
removed anti-ban because it was causing the script to break a lot.


ChrisHemmens:
Wait with this pull request. I need to make another commit, noticed a slight bug.

ChrisHemmens:
Commit pushed. Bug is fixed. Please merge.

PR #1522, opened 2016-10-29
Improve Combat interface; Add more documentation

Moosheplusplus:


Strikegently:
combine this with 
https://github.com/powerbot/powerbot/pull/1525

Moosheplusplus:
this adds functionality to the class and is not just documentation.

src/org/powerbot/script/rt4/Combat.java
```diff
@@ -18,14 +18,40 @@ public int health() {
...
 		return ctx.varpbits.varpbit(301) == 1;
 	}
 
+	/**
+	 * Whether or not the player is in a multi-combat area.
+	 *
+	 * @return <ii>true</ii> if within a multi-combat area, <ii>false</ii> otherwise.
+	 */
+	public boolean inMultiCombat() {
+		return ctx.varpbits.varpbit(1021) == 0x20;
```
ArtificialPB: You should shift and mask the value because only one bit is used to determine if you're in multi-way combat. Other bits could be used for other things, and if they're active at the same time it would break the method.
```
return ctx.varpbits.varpbit(1021, 5, 0x1) == 1;
```

PR #1524, opened 2016-10-29
Grabbing local player's total skill levels

sethrem:
Grabs player's total levels


Strikegently:
Pants-on-head retarded.

sethrem:
Why would you get it by looping through your skills? It's much easier to do it this way as the widget doesn't need to be visible to grab the data from. That's like saying why does the current api get local player's hp by widgets instead of level.

PR #1527, opened 2016-10-30
rt4 Combat overhaul

sethrem:


sethrem:
Accidentally forgot to un-branch my skills update, but I revised the pr so it shouldn't be there anymore.

src/org/powerbot/script/rt4/Combat.java
```diff
@@ -47,4 +67,95 @@ public Boolean call() throws Exception {
 			}
 		}, 300, 6);
 	}
+    
+    	/**
+	 * Gets the wilderness level that you're in
+	 * @return your player's wilderness level, 0 if you're not in the wilderness
+	 */
+	public int wildernessLevel() {
```
rs-sliske: not sure that this belongs in combat

src/org/powerbot/script/rt4/Combat.java
```diff
@@ -47,4 +67,95 @@ public Boolean call() throws Exception {
...
+				return autoRetaliate() == enable;
+			}
+		}, 300, 6);
+	}
+	
+	/**
+	 * Gets the name of your player's weapon
+	 * @return wielded weapon's name
+	 */
+	public String weaponName() {
```
rs-sliske: is it possible to get weapon id / object rather than just the name?

src/org/powerbot/script/rt4/Combat.java
sethrem: I would say getting your wilderness level is indeed combat related.

src/org/powerbot/script/rt4/Combat.java
sethrem: Yes it is but I'd say it's better to do that in equipment instead. Maybe the same with weapon name but it's in the attack tab and doesn't need to be open to grab, so I just put in in there.

rs-sliske: afaik equipment doesnt need to be open to get the data (apart from ammo and ring slot)

if you want to include weapon info in combat class i would just call the equipment method tbh, 

the problem with using item names is they arent unique.

iitc:
- healthRatio should be healthPercent and return a double. Printing to screen and such can truncate and format. Having the precision to say my healthPercent is 95.6% is a lot better than 95%.
 - specialAttack, you should break off the component finding into a method just to avoid the ugly break-loop style. 
 - wildernessLevel, you should check if the component is valid and visible before replacing the text, trimming, etc. (also, the text could be null)
 - You need to store all these component ids and loop increments as constants.

src/org/powerbot/script/rt4/Combat.java
ArtificialPB: This is completely unnecessary. You can get weapon name by parsing equipment.

src/org/powerbot/script/rt4/Combat.java
sethrem: But the wilderness itself is combat related even though you can skill in it. I just find it that this is best suited in the combat class because of that but your point is valid.

src/org/powerbot/script/rt4/Combat.java
```diff
@@ -47,4 +67,95 @@ public Boolean call() throws Exception {
...
+			return 0;
+		}	
+	}
+	
+	/**
+	 * Checks if auto retaliate is on or not
+	 * @return true = on, false = off
+	 */
+	public boolean autoRetaliate() {
+		return ctx.varpbits.varpbit(172) == 0;
```
ArtificialPB: You should mask the value. The same varpbit can store multiple different states, which can all be active at the same time. This means it will return a different value when each of the different states are active. That's why it's important to "extract" only the state you need.
```
ctx.varpbits.varpbit(172, 0, 0x1) == 0;
```

src/org/powerbot/script/rt4/Combat.java
ArtificialPB: I agree with you on that. Let's wait for more feedback about this.

sethrem:
Gonna fix this stuff tomorrow. I've been busy, thanks everyone for the feedback and advice.

src/org/powerbot/script/rt4/Combat.java
Timer: I agree this should be put into the Game class.

src/org/powerbot/script/rt4/Combat.java
Timer: Please follow @ArtificialPB's recommendation.

src/org/powerbot/script/rt4/Combat.java
```diff
@@ -47,4 +67,95 @@ public Boolean call() throws Exception {
...
+	}
+	
+	/**
+	 * Attempts to switch your auto retaliate status
+	 * @param true = enable auto retaliate, false = disable auto retaliate
+	 * @return if the switch has been successful
+	 */
+	public boolean autoRetaliate(final boolean enable) {
```
Timer: If you could follow our code conventions (brackets and finalizing variables that'd be great!). e.g. `final Component c`.

src/org/powerbot/script/rt4/Combat.java
```diff
@@ -47,4 +67,95 @@ public Boolean call() throws Exception {
...
+		if (autoRetaliate() == enable)
+			return true;
+		Component c = ctx.widgets.component(593, 27);
+		return c != null && ctx.game.tab(Tab.ATTACK) && c.visible() && c.click() && Condition.wait(new Callable<Boolean>() {
+			@Override
+			public Boolean call() throws Exception {
```
Timer: Please use Condition.Check, e.g.
```java
		Condition.wait(new Condition.Check() {
			@Override
			public boolean poll() {
				return false;
			}
		});
```

src/org/powerbot/script/rt4/Combat.java
Timer: Please remove this method per discussion above.

src/org/powerbot/script/rt4/Combat.java
```diff
@@ -47,4 +67,95 @@ public Boolean call() throws Exception {
...
+		}
+		return arr.toArray(new String[arr.size()]);
+	}
+	
+	/**
+	 * Gets the selected combat style index
+	 * @return combat style index
+	 */
+	public int styleIndex() {
+		return ctx.varpbits.varpbit(43);
```
Timer: Please heed @ArtificialPB's advice.

src/org/powerbot/script/rt4/Combat.java
```diff
@@ -47,4 +67,95 @@ public Boolean call() throws Exception {
...
+	}
+	
+	/**
+	 * Changes your current attack style to said index
+	 * @param index - index to select (0-3)
+	 * @return index has been successfully changed
+	 */
+	public boolean styleIndex(final int index) {
```
Timer: If you could create and Enum instead of an index for these things, that'd be great. If not we can look into doing it later.
Also please use our code convention in this method and other methods.

sethrem:
@Timer thanks for the feedback, will do I'll update these tomorrow been busy, also I thought you guys didn't like brackets and stuff so I formatted it like that xD.

PR #1543, opened 2016-11-30
compiler errors and version 1.1

armark1ng:
didn't know that powerbot uses compiler 1.6 fixed the errors and now it should work


armark1ng:
im running it on local no problem, 
if it gives any compiler error please lmk

Issue #1519, opened 2016-10-28
Actor.healthPercent() broken rt4


ArtificialPB:
Returns some weird numbers, e.g. 7% when the Npc still has more than half of the health.

Moosheplusplus:
![](https://i.imgur.com/2GxVoSl.png)

can confirm.

ChrisBPB:
@ArtificialPB's solution has worked for over a year so I assume the fix can be considered permanent

PR #1547, opened 2016-12-04
Fixed rt4 Actor#healthPercent()

ArtificialPB:
Solves https://github.com/powerbot/powerbot/issues/1519

[CombatStatusData#getHealthRatio()](https://github.com/powerbot/powerbot/blob/rsbot-api/src/org/powerbot/bot/rt4/client/CombatStatusData.java#L14-L16) returns 30 when health bar is visible and actor has full hp. Tested with both Npcs and Players and it returns the correct value.


Timer:
I'm not sure if this is actually fixing it; this constant changes all the time -- we need to find a hook for it.

ArtificialPB:
Looks like this will be just a temporary solution then.

PR #1554, opened 2016-12-20
Don't deposit inventory if already empty

Moosheplusplus:
Also make use of `Backpack#get()` rather than `Backpack#select()`; internal API calls _**should not**_ be filtering or selecting queries.


Timer:
8a2054d4e6454c6c70eac3bf4af2331f5bd5417d

PR #1557, opened 2016-12-29
Bugfixes in Script

NoodIeking:
Removed several bugs in the script

### DIFF
--- a/src/Noodleking/FishingLobsters.java
+++ b/src/Noodleking/FishingLobsters.java
@@ -1,4 +1,4 @@
-package Noodleking;
+package Fishing;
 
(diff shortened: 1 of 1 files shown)


src/Noodleking/FishingLobsters.java
```diff
@@ -1,4 +1,4 @@
-package Noodleking;
+package Fishing;
```
rs-sliske: dont think this will work. packages are meant to follow the directory structure of your project

NoodIeking: Yeah sorry, the project on my pc has a different package, forgot to change it

PR #1563, opened 2017-01-02
Scripts

NoodIeking:
Added option to fish Swordfish instead of Lobsters (you can choose when starting the script)


Strikegently:
Merged.

In the future, please make your commit comments and Pull Request titles more meaningful.

PR #1585, opened 2017-03-10
CrazeFisher Initial SDN Request.

alexm118:


Strikegently:
1. JForm Designer files won't work on the SDN
2. Don't commit your class files - just your source code

alexm118:
@Strikegently 

Thanks, Removed the class files and the JFormDesigner. Also removed the shop buying script because it is bombing my build for using 1.7 features. Ammended my .gitignore, we may want to have a global gitignore if we don't already.

Strikegently:
why are you removing other people's scripts?

alexm118:
Because it doesn't make sense to have a script that isn't compatible or working on the SDN.

Strikegently:
What 1.7 features was it using?

alexm118:
<> within his GUI to identify types.

Strikegently:
So make a another PR to fix it; don't just remove his scripts.

alexm118:
Why should I be fixing someone elses scripts? I wouldn't want or expect anyone else to be fixing my scripts and we shouldn't be hosting broken scripts on the SDN.

Strikegently:
>Why should I be fixing someone elses scripts? 
That's like 30% of the reason of open source software.

Strikegently:
If you're going to modify someone else's script, then do so in a separate PR.

PR #1594, opened 2017-04-05
RT4 CacheItemConfig new opcode

ArtificialPB:
A new opcode 42 was added recently.


ArtificialPB:
Opcode 42 is shift action index, added shift action loading.

PR #1597, opened 2017-04-12
Cancel current action before interacting with random event

### LINKED ISSUE #1593: Random Event freezes if script is in middle of selecting item/casting spell
It seems to be an increasing issue on all of my scripts; The random event handler will take over control, be in the middle of something (i.e. casting spell, using item, etc), and will cause it to not escape and completely break the script.

Moosheplusplus:
Fixes #1593.

### DIFF
--- a/src/org/powerbot/bot/rt4/RandomEvents.java
+++ b/src/org/powerbot/bot/rt4/RandomEvents.java
@@ -14,14 +14,14 @@ public class RandomEvents extends PollingScript<ClientContext> {
 	}
 
 	private boolean isValid() {
-		return ctx.input.blocking() && !ctx.properties.getProperty("randomevents.disable", "").equals("true")
-				&& !ctx.npcs.select().within(5d).action("Dismiss").select(new Filter<Npc>() {
-			@Override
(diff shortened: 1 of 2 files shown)


src/org/powerbot/bot/rt4/RandomEvents.java
```diff
@@ -32,8 +34,16 @@ public void poll() {
...
 		}
-		final Npc npc;
-		if ((npc = ctx.npcs.poll()).interact(false, "Dismiss")) {
+		final Npc npc = ctx.npcs.poll();
+		if(ctx.input.move(npc.nextPoint())) {
+			for(MenuCommand a : ctx.menu.commands()) {
+				if(!a.option.contains(" -> "))
+					continue;
+
+				ctx.widgets.component(Constants.CHAT_WIDGET, Constants.CHAT_VIEWPORT).click();
```
Timer: How do we know this is *always* safe to click? Can we choose something more appropriate?

Moosheplusplus: As far as I know, there are no spells you can cast or items you can use on chat messages.

Issue #1419, opened 2016-05-01
Force left click ctx.bank.open()


Zarama:
Could you possibly overload `Bank.open()` function to left click bank?
There are times when we are standing at places where the bank is behind mobs and then makes it right click bank. 
This just causes delays and lags.

`Bank.open(boolean left_click)`

Strikegently:
>There are times when we are standing at places where the bank is behind mobs and then makes it right click bank.

Well no shit. If the bank is behind another entity, it needs  to right click so it can open the bank.

Zarama:
Not really, it can still left click. 
That's why we should allow for the overloaded function, so we can left click. 
A human wouldn't right click on a bank chest when they can clearly see it and click it.

ArtificialPB:
@Strikegently The problem sometimes arises if there are too many people standing beside the bank. Index of bank opening menu action wont be the first one and it's usually off-screen. This causes the bot to get stuck in a continuous loop right-clicking the bank while it can be opened by left-clicking it.

I've solved this in my RS3 scripts by checking if the object can be right-clicked by calculating if the menu index will be off-screen or not and then checking if the tooltip while hovering the object is the same as the desired action (in this case it would be "Open"). If tooltip action and option match the ones provided, do a left click.

Timer:
The menu needs to be fixed before this can be done.

Timer:
dependent on #1032

PR #1598, opened 2017-04-12
Improved rt6 interactions, added tooltip menu command and some documentation

ArtificialPB:
Sometimes there are a lot of players in front of the entity you're trying to interact with and the client gets stuck in a continuous loop if the menu command is off-screen. You can still left-click it in this case.

This update checks the tooltip if you can still left-click the entity or not and fixes https://github.com/powerbot/powerbot/issues/1419.

### DIFF
--- a/src/org/powerbot/script/rt6/Menu.java
+++ b/src/org/powerbot/script/rt6/Menu.java
@@ -22,6 +22,8 @@ import org.powerbot.script.Vector2;
  * Utilities pertaining to the in-game menu.
  */
 public class Menu extends ClientAccessor {
+	private Component tooltipComp = null;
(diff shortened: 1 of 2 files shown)


src/org/powerbot/script/rt6/Menu.java
```diff
@@ -22,6 +22,8 @@
  * Utilities pertaining to the in-game menu.
  */
 public class Menu extends ClientAccessor {
+	private Component tooltip_comp = null;
```
Strikegently: Java naming conventions, please.

ArtificialPB:
@Strikegently Done.

Issue #1592, opened 2017-04-03
rt4 Item#nextPoint() seems to be off


Moosheplusplus:
The deviation is too high and the mean seems to be off from `Item#centerPoint()`

![](https://i.imgur.com/WJte0ZN.png)

Moosheplusplus:
@ArtificialPB @Paris @Strikegently the deviation is still too high to the point where it misclicks ~1 out of 5 times.

![](https://i.imgur.com/dDJ0JKu.png)

```
[INFO] Samples: 17000, Good: 13871, Bad: 3129, Percentage wrong: 18.406%
```

PR #1599, opened 2017-04-12
Fixed rt4 Item#nextPoint()

ArtificialPB:
Fixes the mean in next point calculation. Center rectangle must be offset for 1/8 of the bounding one in order for its center point to be the actual item center point. Fixes https://github.com/powerbot/powerbot/issues/1592, @Moosheplusplus the deviation looks fine.


Strikegently:
@Moosheplusplus 
Review when you get a chance.

Timer:
I assume this is because of broken component location, not an actual logic issue?

ArtificialPB:
@Timer OSRS doesn't have any component issues. This is a logic one.

PR #1608, opened 2017-04-21
zPKHelper initial release

ElZaiat:
Thread: https://www.powerbot.org/community/topic/1331392-osrs-zpkhelper/


Strikegently:
This seems more like an OSBuddy plugin than an RSBot script.

ElZaiat:
Does that seem to be a problem? Anyway, I will be adding features like Auto K.O and Auto Eat in the future.

Strikegently:
PKing a competitive event in RuneScape. Features like those seem like they would give the user an unfair advantage against a human.

ElZaiat:
Isn't the bot's purpose to give users unfair advantage 0.o?

Strikegently:
Not in direct competition with other players, no.

Issue #1611, opened 2017-04-25
Upgrade antipatterns


ArtificialPB:
Current client antipatterns don't do much and they might disrupt the execution of scripts. I propose the addition of non-interfering antipatterns. Each user would get a randomly generated pattern profile which would be stored on the server (or just locally on their computer) and loaded when they start the script. The pattern profile could also be slightly adjusted between runs for things like fatigue, the frequency of certain actions, clicking accuracy, etc... . 

Here are a few "patterns" that could be implemented directly into the client:
- Closing interfaces with mouse (clicking the "X") or keyboard (pressing ESC button).
- Clicking accuracy (next point calculation deviation).
- Continuing chat with mouse or keyboard.
- Energy limit for turning on running.

Script writers would also have the ability to store their own patterns in the profile. This would make writing custom antipatterns much easier because right now we have to override client methods/write our own custom implementations to add things like the ones written above, which could be added into the client by adding one or two new lines of code. And many script writers don't even implement any antipatterns so this would automatically add a base layer of protection.

## Phase one
- [ ] Implement Break Handler
- [x] Add universal interface closing method (#1647)
- [ ] Implement Antipatterns API
  - [ ] Save/load patterns
  - [ ] Option to clear and recalculate new patterns
- [ ] Overloaded closing methods with hotkey support
  - [x] Deposit boxes (rt4 & rt6) (#1670)
- [x] Add hotkey support for game tab switching (rt4)
  - [x] Add keybinds API (#1620)
  - [x] Implement hotkey support (#1620)
- [ ] Add RS3 hotkey support (#1639)
- [ ] Add support for camera rotations with mouse
- [ ] Implement additional game settings checks
  - [x] ESC closing enabled (rt4) (#1620)
- [ ] Add in-game world switching support rt6
- [x] Mouse scrolling support rt4 (#1622)
- [ ] Add Hud menu/window/subtab support for all game mode combinations (legacy combat & eoc interface, and vice versa)

MrFreezerburnt:
Closing interfaces with ESC requires a certain option toggled to work (for rt4 at least, I couldn't find a toggle for rt6 but it works). Should devs consider this, check if they have the option selected. Here's the varpbit for rt4:

`1224 - 1145077760 (0x44408000) -> -1002405888 (0xc4408000)`

ArtificialPB:
[...]

PR #1614, opened 2017-04-29
added hotkey option to close bank

Strikegently:
- adds a `bank.close(boolean)` method, which allows the user to specify whether to use the hotkey or the mouse.
- keeps the default `bank.close()` as an overloaded method; returns `bank.close(false)` by default to keep all script behaviors the same.

### DIFF
--- a/src/main/java/org/powerbot/script/rt6/Bank.java
+++ b/src/main/java/org/powerbot/script/rt6/Bank.java
@@ -186,13 +186,20 @@ public class Bank extends ItemQuery<Item> implements Viewable {
 		return opened();
 	}
 
+	public boolean close() {
+		return close(false);
+	}
+
 	/**
-	 * Closes the bank by the 'X'.
+	 * Closes the bank. If true, uses the hotkey (ESC). If false, clicks 'X'.
 	 *
 	 * @return <tt>true</tt> if the bank was closed; otherwise <tt>false</tt>
 	 */
-	public boolean close() {
-		return !opened() || ctx.widgets.component(Constants.BANK_WIDGET, Constants.BANK_CLOSE).click("Close") && Condition.wait(new Condition.Check() {
+	public boolean close(final boolean hotkey) {
+		return !opened()
+				|| ((hotkey && ctx.input.send(Constants.HOTKEY_CLOSE))
+				|| ctx.widgets.component(Constants.BANK_WIDGET, Constants.BANK_CLOSE).click("Close"))
+				&& Condition.wait(new Condition.Check() {
 			@Override
 			public boolean poll() {
 				return !opened();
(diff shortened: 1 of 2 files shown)


Strikegently:
@Timer @ArtificialPB @k0na
Should the default method use the hotkey or the mouse?
Should there be a random distribution between the two as part of an "antipattern"?

ArtificialPB:
I'd say use the mouse for now. We'll add randomization when the antipatterns API gets added so we can save and load patterns for consistency.

k0na:
I agree with @ArtificialPB, for now anyway. 

There are a variety of factors that influence whether players use hotkey(s) / mouse clicks to perform actions - none of which the API currently takes into account. To name a few: in-game activity, computer type (desktop/laptop), fatigue, window size etc. Once there is a more robust antipattern framework, things like these should all weigh in on decisions like this one to help select the smarter interaction type.

Paris:
I've merged the commit but I agree with the other comments, decisions like this should be part of the bot not left to script writers. For now it's determined randomly.

rs-sliske:
imo the issue with the bot deciding itself which to use means there is know way of knowing what the script is doing around banking

tbh the only place where explicitly closing the bank is optimal is when you are staying next to the bank (skills such as crafting and herblore, and a few lunar spells)

ArtificialPB:
@rs-sliske There are no methods to close the bank by clicking on map, so I don't know how implementing this makes any difference. Either way you have to write your own method to do that. 

You could make a PR with this addition, falling back to the default method if you're unable to click on map.

rs-sliske:
changing the default from clicking close button to pressing escape feels like its making problem worse

surely the goal is to try and make scripts behave more like legit players, on osrs at least, i dont think closing interfaces with escape even works unless you go in the options interface and enable it

ArtificialPB:
In my opinion the current implementation will make no change, because everyone was and will still use the same pattern over time, just the pattern is now different.

We're aware of the fact that some functions can be enabled/disabled in-game. The client will handle this accordingly.

How many script writers have implemented their own antipatterns? And I'm not talking about random camera/mouse movements, but randomizing things like bank closing for each user and saving it for persistency between script runs? This is what we're trying to achieve here, persistent (across different scripts and runs) and slightly randomized pattern for each user.

If you have any other concerns over this, please post in #1611 so we have everything in one place.

rs-sliske:
whether the new pattern is any better or worse on its own doesnt really make much difference

the problem with changing the default behavior of the client is that the pattern changes for every bot at the same time (which itself feels like a pretty obvious pattern, especially given that the new pattern isnt really any more efficient)

looking at the commits that were made, it doesnt look like there is any checking for if the user has the option enabled, or remembering whether the user is going to use escape/mouse for closing bank. it is just calling `Random.nextBoolean` every time close gets called

even the parameter coma added to allow script writers to implement something like that themselves was removed by paris

PR #1620, opened 2017-05-13
Added rt4 keybind support

ArtificialPB:
Contains a check if ESC can be used for interface closing.

### DIFF
--- a/src/main/java/org/powerbot/script/rt4/Game.java
+++ b/src/main/java/org/powerbot/script/rt4/Game.java
@@ -31,25 +31,47 @@ public class Game extends ClientAccessor {
 	}
 
 	/**
-	 * Attempts to open the specified game tab on the user interface. If the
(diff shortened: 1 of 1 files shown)


ArtificialPB:
Tabs can now be opened with hotkeys, moved Keybind enum to the Game class.

Strikegently:
Pinging @Moosheplusplus 
I don't play OSRS.

rs-sliske:
@Strikegently neither does mooshe :p 

not had chance to check all the varpbit settings are correct etc, but overall process looks reasonable. osrs allows users to choose which f-key opens each interface. and as long as this doesnt end up with another dice roll i think it would be a useful thing to have

Moosheplusplus:
@rs-sliske I imagine he pinged me because I previously proposed keybind support in the past on the forum, but I never submitted a PR.

PR #1622, opened 2017-05-13
Added support for mouse and bar scrolling rt4

ArtificialPB:
Deprecated original method. The new implementation includes some small delays to better mimick players looking for a component, and actually has the option to move the scroll bar by clicking rather than scrolling.


Strikegently:
If you're going to deprecate the old method, make it return the new method so that the old one does not need to be maintained (but will stay as to not break anything).

ArtificialPB:
It was left it as it was because the functionality was different.

As per your request, I've made the old, deprecated method return the new one.

Issue #1544, opened 2016-12-02
Component location and visability/validity hooks broken


ArtificialPB:
This happens in rt6 and it started happening after one of the recent game updates:

![component issue](https://cloud.githubusercontent.com/assets/4672155/20845597/20c77ae0-b8c6-11e6-9a5d-c9f296c50632.png)

Timer:
Broken hook!

PorkScripts:
Is it just due to updated components?

For example: When in game, the recent chat box messages are in widget 137, component 133. My RSBot client is reporting the absolute coordinates of this component at (2, 25) which are wrong.
The correct coordinates can be found by finding the chat box container in widget 1477, component 102. Its absolute coordinates are (4, 434). Add these to the incorrect coordinates above (2, 25) and you get (6, 459) which should be correct.

So is RSBot depending on widget 1477 for offsetting other widgets? And maybe the components in 1477 were updated from a game update, screwing up offsets of certain interfaces?

Also, another issue I've found it the options menu (widget 1433) is showing visible as "true" even after closing it.

Pls fix thx

ArtificialPB:
In addition to location offset, some widgets and components are always valid/visible, even when closed. One of those interfaces for example is Options Menu:
![componentissue2](https://cloud.githubusercontent.com/assets/4672155/21179417/42f6186a-c1f4-11e6-9c62-bab32f5cdac2.png)

scriptnoob2000:
Any ideas when this will be fixed? I'm also experiencing this issue and it blocks me from continuing work on my artisan blacksmithing script since all mini game components are also broken with the same issues 1) Always visible 2) Position on screen is reported wrongly.

ArtificialPB:
Please make this a high-priority fix. A lot of RS3 script are at least partially broken due to broken hook.

Timer:
d35f5564e568cfe885cf7fbf96efcf0972128f59

ArtificialPB:
Still happens, including with always returning valid/visible for some components. 
<img width="609" alt="screen shot 2016-12-24 at 18 28 21" src="https://cloud.githubusercontent.com/assets/4672155/21468181/dd49582c-ca06-11e6-8544-1e9dde2d7aeb.png">

k0na:
This is still broken @Timer

Strikegently:
Hud doesn't work at all. Widgets return valid and visible even when closed. 
My scripts are almost entirely non-functional.

k0na:
[...]

PR #1624, opened 2017-05-17
Potential components fix

ArtificialPB:
A potential fix addressing #1544. Raising the recursion depth seems to fix the issue for most, if not all, components.

### DIFF
--- a/src/main/java/org/powerbot/script/rt6/Component.java
+++ b/src/main/java/org/powerbot/script/rt6/Component.java
@@ -180,7 +180,8 @@ public class Component extends Interactive implements Drawable, Displayable, Ide
 			if (uid == node.getUid()) {
 				return (int) node.getId();
 			}
-			if (i++ >= RECURSION_DEPTH) {
+			if (i++ >= 1500) {
+				System.out.printf("WARNING: parentId operation killed -- beyond depth of %d.%n", 1500);
 				break;
 			}
 		}


Strikegently:
@Paris 
Please don't merge this until I've had a chance to review it. I'll take a look after work tomorrow.

@ArtificialPB 
Were you able to verify that positions, locations, and validity are functional without the workarounds?

ArtificialPB:
I focused on the parentId value, it was -1 at some point in the offset components. I haven't explicitly tested with locations and the workarounds. I'll take a look at it again today. You can check for yourself, parentId for 1371, 10 is always returning -1 before implementing this. I've compared the value to the one in the cache and it matches. Needs to be verified if the recursion depth is high enough for every case.

ArtificialPB:
Tested and it fixes the locations, positions, and validity without the workarounds. Recursion depth seems okay, there were still some components returning `-1` for parentId, but that was happening even after removing the recursion limit.

Components whose parentId was still `-1` (42 in total when just standing in the game without any additional interfaces opened) seem to fallback to `107738123` in the [cache](https://github.com/ArtificialPB/DevelopmentKit/blob/master/src/com/artificial/cachereader/wrappers/rt6/Component.java#L134). It looks like the client already handles those cases [here](https://github.com/powerbot/powerbot/blob/rsbot-api/src/main/java/org/powerbot/script/rt6/Component.java#L222-L228) (needs additional verification - @Timer).

Issue #971, opened 2015-10-04
Component tooltips


Strikegently:
A method of getting the tooltip text associated with a component would be fantastic. 
Nearly 100% of components could be found dynamically, which would prevent scripts from ever being broken from widgets again.

![](http://i.imgur.com/laxzZ2o.png)

Strikegently:
The component for all of these tooltips is 1486, 26 if you wanted to add a "getCurrentTooltip" method.

PR #1626, opened 2017-05-24
Added rt6 component tooltip support

ArtificialPB:
As requested in https://github.com/powerbot/powerbot/issues/971. Component#tooltip() already existed so I added a method named componentTooltips() to avoid confusion.


Strikegently:
@Timer

PR #1632, opened 2017-05-30
Added new opcodes to rt4 cache reading

ArtificialPB:
Parameters loading was added to rt4 cache definitions with a recent game update. This PR also adds some missing opcodes and adds item equip actions decoding.


Strikegently:
@Timer

PR #1656, opened 2017-06-13
removed multi trycatch

talahkz:


Strikegently:
Desktop will be blocked by the Security Manager.

Use AbstractScript.openUrl()

PR #1665, opened 2017-06-19
Scripts

TendingInfinite:
Solved some fishing issues


Strikegently:
In the future, please give more detailed commit messages and PR titles; it's also incredibly useful if you branch, so it ensures clean merges.

PR #1666, opened 2017-06-19
Hoping I did everything correctly

Kyle628:


src/powerkyle628/scripts/LogIncinerator.java
```diff
@@ -0,0 +1,154 @@
...
+
+    final static int normaLogId = 1511;
+    final static int willowLogId = 1519;
+    final static int yewLogId = 1515;
+    final static int oakLogId = 1521;
+    final static int mapleLogId = 1517;
+    final static int mahoganyLogId = 6332;
+    final static int teakLogId = 6333;
+    final static int magicLogId = 1513;
+    final static int elderLogId = 29556;
```
MrFreezerburnt: You're referencing rt4 for OS, which matches your manifest. However, elder logs only exist in RS3 (rt6). Either use rt6 or change the inconsistencies.

src/powerkyle628/scripts/LogIncinerator.java
```diff
@@ -0,0 +1,154 @@
...
+    @Override
+    public void poll() {
+        if (getLogCount() > 0 && !inBank()) {
+            burn();
+        } else if (!inBank()){
+            walkToBank();
+
+        } else if (inBank() && getLogCount() > 0) {
+            leaveBank();
+        } else {
```
MrFreezerburnt: Having an else statement in your poll() method means that your script will always be doing something, even when you don't want it to. What happens if the script gets stuck or none of your methods return the proper values?

src/powerkyle628/scripts/LogIncinerator.java
```diff
@@ -0,0 +1,154 @@
...
+        Condition.wait(new Callable<Boolean>() {
+            @Override
+            public Boolean call() throws Exception {
+                return !inBank();
+            }
+        });
+    }
+
+    public boolean inBank() {
+        Area bankArea = new Area(new Tile(3179, 3447), new Tile(3191,3433));
```
MrFreezerburnt: Make your bankArea a class-wide variable and have it populated once your script starts. Currently, your poll() method has the possibility of creating the same bankArea three times, which is a waste of resources and could potentially slow down the script.

src/powerkyle628/scripts/LogIncinerator.java
```diff
@@ -0,0 +1,154 @@
...
+
+
+    public void burn() {
+        //how many logs we start with could be useful
+        int ogLogAmount = ctx.inventory.select().id(logChoiceID).count();
+
+        Item tinder = ctx.inventory.select().id(tinder_id).poll();
+        Item log = ctx.inventory.select().id(logChoiceID).poll();
+
+        log.interact("Use");
```
MrFreezerburnt: What happens if the game lags, or disconnects? For example, this method will click on the log, instantly snap to the tinderbox, click it, then you wait for the animation to finish (twice...?). You need to add conditional waits, like you have below.

src/powerkyle628/scripts/LogIncinerator.java
```diff
@@ -0,0 +1,154 @@
...
+                return ctx.players.local().animation() == -1;
+            }
+        });
+        //check to see if we successfully burned anything
+        int newLogAmount = ctx.inventory.select().id(logChoiceID).count();
+        if (ogLogAmount == newLogAmount) {
+            //must have been standing on something, move somewhere random nearby
+            int randomX = org.powerbot.script.Random.nextInt(0,5);
+            int randomY = org.powerbot.script.Random.nextInt(-2, 2);
+            Tile randomTile = ctx.players.local().tile().derive(randomX,randomY);
```
MrFreezerburnt: As it stands, this method will always attempt to light a fire once, then move to a randomly derived tile. You never want to derive random numbers to handle a situation like this. What happens if your method makes the player walk into a bank? There's a random chance they could never leave, and your script will try to light a log anyway each time the method is called, making this look EXTREMELY bot-like.

src/powerkyle628/scripts/LogIncinerator.java
```diff
@@ -0,0 +1,154 @@
...
+            ctx.movement.step(randomTile);
+        }
+
+
+    }
+
+    public void walkToBank() {
```
MrFreezerburnt: See above about conditional waits.

src/powerkyle628/scripts/LogIncinerator.java
```diff
@@ -0,0 +1,154 @@
...
+        System.out.println("walking");
+        final Locatable nearestBank = ctx.bank.nearest();
+        ctx.movement.step(nearestBank);
+    }
+
+    public void bank() {
```
MrFreezerburnt: See above about conditional waits.

MrFreezerburnt:
Your script could use some work, and could potentially be converted into a Task-based script. https://www.powerbot.org/community/topic/1182786-the-rsbot-cookbook-writing-your-first-script-using-tasks/

Kyle628:
Thanks for such a thorough review. I definitely didn't think of some of the stuff you pointed out. I think I can make all of the changes easily, except for the part where I have the bot move to a random nearby tile. The reason I did this is because if the bot was standing on top of an already lit fire, it would just keep failing to light logs on top of it. What is the proper way to handle this? I felt like my solution was alright, because if the bot ends up in the bank, leaveBank() will get called because getLogCount() will return > 0

ChrisBPB:
There are a lot of ways to approach that, but moving randomly can get you in issues and perhaps even stuck. The way I did it was to have designated start points (like most humans do anyway, no one just places fires randomly) then count fires on each of the rows of my starting points and start lighting at the most empty one 
`ctx.objects.select().select(new Filter<GameObject>() {
                @Override
                public boolean accept(GameObject go) {
                    if (go.tile().y() == startTiles[index].y() && go.name().equals("Fire")) {
                        return true;
                    }
                    return false;
                }
            }).size();`


but also you need to perhaps check again incase someone is torlling you and lighting fires infront of you and recalculate again or even world hop

Kyle628:
I made the changes, and the bot is looking a lot smoother now...thank you!

PR #1671, opened 2017-06-24
Fix random event dismissal

ChrisBPB:
random events can not be dismissed by the client if they occur when an
inventory item is selected, this will deselect the item if one is
selected


Strikegently:
- fix spacing
- when submitting pull requests, it is conventional to submit them on a separate branch (we haven't adhered to this, but we really should).

Strikegently:
Please fix the spacing.

ChrisBPB:
the space infront of if?

Strikegently:
the space between the values and the operator. I fixed it.

PR #1672, opened 2017-06-26
Revert "Added rt6 component tooltip support"

Strikegently:
Reverts powerbot/powerbot#1626

@ArtificialPB 
This does not build.
- missing imports
- unknown method `id()`


ArtificialPB:
Sucks not being able to build locally. I'll get it fixed.

PR #1674, opened 2017-06-28
Rsbotupdate

Kyle628:


Strikegently:
Few things:
- you don't need to call `ctx.controller.stop()` in your `stop` method. This method is only called when you stop the script, so stopping the script when it's stopping the script is unnecessary.
- Why are you clicking exact pixels in your `stop` method?

Kyle628:
Oops, I will take that out.

I am clicking the exact pixels, because I first tried to logout using widgets, but I couldn't figure out how to click the "door" button. It doesn't have text so I couldn't find it in the widget explorer. I googled around, but couldn't find a logout method anywhere.

Strikegently:
did you try `game.logout()`

Kyle628:
In the docs I can only find game.loggedin() which just returns whether you or logged in or not.

ctx.game.logout();

throws me a cannot resolve method error

rs-sliske:
`ctx.game.tab(Tab.LOGOUT)` should switch to that tab

Kyle628:
Ooooh nice, that should help!

Kyle628:
Okay, looks much better now using tabs and widgets

PR #1677, opened 2017-07-03
Scripts

talahkz:


Strikegently:
- Your PRs and commit messages should be more detailed than "asd"
- PRs should serve one purpose; i.e. don't update one script and add another in the same PR
- PRs should originate from a separate branch to ensure clean merges

PR #1680, opened 2017-07-09
amended package declarations

talahkz:


Strikegently:
Pull Requests should originate from separate branches to ensure clean merges.

Also, you don't need to do a commit for each file changed. You could have done a single commit with the message "amended package declarations" for all 3 files.

PR #1698, opened 2017-07-30
update script

weylen:
script stopped change to script suspend when the condition start starts.
add status when script suspend.


Strikegently:
Your commits and pull request titles should be more descriptive.

PR #1708, opened 2017-08-11
Adding FreeRunecrafter and JangerBerryGrabber

needscroll:
Threads here
https://www.powerbot.org/community/topic/1336578-500khrs3-jangerberry-grabber/
https://www.powerbot.org/community/topic/1336214-freers3-freerunecrafter/


Strikegently:
I would prefer that each pull request serve a single purpose, so do two separate pull requests for adding two separate scripts in the future.

PR #1721, opened 2017-09-13
first upload SalmonCollector

turningwind:


Strikegently:
Please rename your main script class something other than "Main".

turningwind:
Changed it, may I know the reason for this?

Strikegently:
This won't compile. The file is named SalmonCollector but the class is still named Main.

>Changed it, may I know the reason for this?
Scripts on the SDN are identified by user ID and main script class name. All Open SDN™ scripts are uploaded to a sandbox account; naming the main class something incredibly generic like "Main" increases the likelihood of conflicts.

turningwind:
I understand, I will make sure to keep that in mind for my next scripts. Thanks for explaining.

PR #1724, opened 2017-09-16
ChocoCrusher

turningwind:


Strikegently:
You need a topic ID. The SDN won't show a script if it doesn't have a topic ID.

turningwind:
Alright, added the topic id's to my scripts.

PR #1738, opened 2017-10-03
Remove deprecations in magic class

ChrisBPB:
spell() is useful and shouldn't be deprecated without replacement, spell relies on component() so that should not be deprecated, and the int widget has a note saying not to remove it so I guess should not be deprecated either


ChrisBPB:
I messed this up cause I didn't branch properly

PR #1739, opened 2017-10-20
SalmonCollector script, with userprofile and java 8

turningwind:
https://www.powerbot.org/community/topic/1337839-free-osrs-f2p-salmoncollector/


Strikegently:
Ideally each pull request should serve a different purpose, rather than one doing multiple things.

PR #1741, opened 2017-10-31
Update World.java

ChrisBPB:
true parameter messed up, passed container and component in wrong order and can't scroll. Same issue with banking previously which was fixed


M0tionl3ss:
Is it possible to add a parameter to the hop method so we can choose to scroll with mouse or to use the ingame scroll bar?

PR #1758, opened 2017-11-21
Update Prayer.java

### LINKED ISSUE #1657: Quick Prayer Setup
https://github.com/powerbot/powerbot/blob/rsbot-api/src/main/java/org/powerbot/script/rt4/Prayer.java#L75

It is failing to interact and this line (L#75) is constantly returning true.

ChrisBPB:
fixes #1657 

must be something weird about the string as "Setup Quick-prayers" does not match it, but just Setup works anyway as expected.


Strikegently:
`Quick-Prayers` may be the target, not the action. In which case, it should be:
`interact("Setup", "Quick-Prayers")`

ChrisBPB:
Yep, is as you said. Have updated it.

PR #1759, opened 2017-11-21
Fix world hopping

### LINKED ISSUE #1757: World hopping bug
When hopping to world 65 there is a warning in the chatbox causing the script to freeze. 
[https://i.imgur.com/7Y1I027.png](https://i.imgur.com/7Y1I027.png)
When adding parameters to the select method it will not give a fresh data set. When all worlds are visited the size of the collection will be 0. Current work around is using a select without parameters to refresh the data set. Bot will also try to hop to King of skill worlds.

ChrisBPB:
Current implementation does not consider the chat you have to answer to enter PVP and high risk worlds, and also all worlds if the user hasn't disabled it. This may have been by choice to stop people entering PVP worlds by accident, but it should be up to the develop to select the right world rather than for the client to not support it by default.

Unsure if we should have a wait condition before checking ctx.chat.canContinue() but I believe this would work as is.

would fix #1757

### DIFF
--- a/src/main/java/org/powerbot/script/rt4/World.java
+++ b/src/main/java/org/powerbot/script/rt4/World.java
@@ -10,6 +10,10 @@ public class World extends ClientAccessor
 
 	public static final World NIL = new World(null, -1, Integer.MAX_VALUE,
 			Type.UNKNOWN, Server.RUNE_SCAPE, Specialty.NONE);
+			
+	public static final int WORLD_SWITCH_PARENT = 219;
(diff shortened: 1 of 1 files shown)


Strikegently:
Could the chat widget be found dynamically, or could it make use of the Chat API?

ChrisBPB:
I haven't really used the chat API but looking at it, it can probably utilise continueChat so I will look into this tomorrow

ChrisBPB:
updated thanks

PR #1765, opened 2017-12-03
Scripts Upload

iDznPB:
Uploading scripts, updated unnecessary files & client ID.


Strikegently:
This needs some work.

One of things that powerbot prides itself on is getting fledgling developers to learn and grow in their knowledge and skills - and this PR shows you need a little guidance.

We all the know the Golden Rule - do to others what you want them to do to you; the Golden Rule in software development should be "if you're writing the same code over and over, you're doing something wrong."

In your NoobNester class, you're writing the same code over and over and over (and over x30), and you had to create dozens of variables to facilitate that. This setup would be incredibly difficult to maintain, since if anything changed with those methods, you would need to make changes to 30 methods which is costly and error prone.

The first thing you need to study is loops. You're doing the same thing several times in a row; this can, and should, be expressed as a loop rather than a continuous stream of statements.

Example:
```
ge = new org.powerbot.script.rt4.GeItem(5075); nestPrice = ge.price;
ge = new org.powerbot.script.rt4.GeItem(7416); ClawPrice = ge.price;
ge = new org.powerbot.script.rt4.GeItem(7416); SkinPrice = ge.price;
ge = new org.powerbot.script.rt4.GeItem(5302); LantadymePrice = ge.price;
...
```

vs.

```
int[] IDs = { 5075, 7416, 5302 };
for( int id : ids ) {
    //do something with new GeItem( id ).price;
}
```

The next thing I would encourage you to study is basic data structures. The most obvious solution to your problem would be a key-value pair structure, which is called a Map in Java. Instead of, again, several repeated statements and variables, you would store your values in a structure for easier use.

Example:

```
nestCount = ctx.inventory.select().id(5075).count();
LantadymeCount = ctx.inventory.select().id(5302).count();
DwarfWeedCount = ctx.inventory.select().id(5303).count();
```

vs.

```
Map<Integer, Integer> itemCounts = new HashMap<Integer, Integer>();
for( int id : ids ) {
    itemCounts.put( id, ctx.inventory.select().id( id ).count();
}
```

There's more, but these are two biggest things. If you have questions, you can post in Developers on the forums or in our Discord. 

P.S.: Don't close this PR. You can make the changes and commit them to this branch, and he PR will automatically update to show the changes.

iDznPB:
Thank you so much for this, I really do appreciate it. This was the first script I made and I was lucky to get it to run. However when it did run due to the sheer amount of variables it lagged significantly when starting the script.

I will at some point redo the nesting script as above to make it much more efficient but I'm definitely trying to learn as much as I can so as I said, it's much appreciated.

iDznPB:
As mentioned on the commit I've updated the variables into a loop and map. Although now I need to be able to check my starting inventory count (in case people start with items in their inventory) VS the difference in inventory after the nest is opened, then run the difference gained VS the price to work out Gross income then minus the expenses. but when I try to refer to the itemCount in any way I return a null pointer error. How would I go about referring to the map?

iDznPB:
Removed NoobNester, if the remaining OgreCannon can be merged, would be appreciated.

PR #1767, opened 2017-12-06
Add RS3 BarbarianFisherAndCooker script

RizaneEves:
This is my first script and would like to add it to the open script repository. This uses TreeBot, a tree-like bot flow (for tree based task creation) that I have made a basic implementation of based on other bot APIs I've used. Re-opening this pull request after fixing bugs and adding a UI.

### DIFF
--- a/src/sheel/BarbarianFisherAndCooker/BarbarianFisherAndCooker.java
+++ b/src/sheel/BarbarianFisherAndCooker/BarbarianFisherAndCooker.java
@@ -1,0 +1,21 @@
+package src.sheel.BarbarianFisherAndCooker;
+
+import src.sheel.BarbarianFisherAndCooker.branches.IsInventoryFull;
+import src.sheel.BarbarianFisherAndCooker.TreeBot.*;
+import org.powerbot.script.Script;
+
+@Script.Manifest(
+        name = "Barbarian Fisher and Cooker",
+        description = "Fishes and cooks at the barbarian village and banks after!",
+        properties = "author=sheel;topic=1340719;client=6"
+)
+public class BarbarianFisherAndCooker extends TreeBot
+{
+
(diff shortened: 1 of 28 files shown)


Strikegently:
Is this ready to merge now?

Also, in the future, please wait to make a PR until you ready for it to be merged.

RizaneEves:
Yes, I apologize for that. It's ready to be merged. Will I have to make another merge if I want to make updates to it (based on feedback from the community)?

ChrisBPB:
Compiler:	package com.sun.xml.internal.bind.v2.runtime.reflect.opt does not exist

RizaneEves:
Hmm... interesting... I'm not getting that error, and not sure how it originates. I'm not ]using that package in any of my scripts. Does it give information about class and line #?

rs-sliske:
its being imported in your main script class but doesnt look like it is actually being used

RizaneEves:
Oh, I see, thanks for pointing that out!

RizaneEves:
@ChrisBPB Could you remove that import line from BarbarianFisherAndCooker.java? Or should I make another merge request?

ChrisBPB:
You should make a new merge request but all your imports aren't how they should be - but it should compile. Your package should just be sheel, not src.sheel (just because it's in the src folder doesn't mean it requires src to be part of the package).

so for example where you write

package src.sheel.barbfisher or whatever should just be
package sheel.barbfisher

and that's the same for all your classes, you shouldn't need src

and then imports would have to change too to drop the src. part

RizaneEves:
Hmm, I see. It's strange though - because if I remove 'src' from my packages/imports in IntelliJ I get a compile error. Is there a setting somewhere I need to change in IntelliJ to fix that?

ChrisBPB:
if you do it like this it should be fine
https://i.imgur.com/joXuz3d.png

you will need to make the changes to all of your classes though

PR #1770, opened 2017-12-10
Fix directory structure and packages

RizaneEves:
Fixing directory structure so new scripts can be added. Package structure also fixed so code will compile on SDN


RizaneEves:
Ah, right. I'll make another PR. That should fix it for sure... Seems like I just messed up my entire project structure, so I just checked out all the code again and moved my files around.  I apologize for that

ChrisBPB:
no worries I just noticed it

PR #1774, opened 2017-12-12
Rewrote withdrawing methods

iDznPB:
Updated the withdrawing method per advice on previous PR, makes it much more efficient using the food as a parameter rather than a new class for each type of withdrawal.

Thanks for the advice and apologies on making another request so soon.


Strikegently:
Reusable code is a beautiful thing.

PR #1777, opened 2017-12-15
Fix directory structure + add OSRS JugFiller

RizaneEves:


ChrisBPB:
There are merge conflicts so we can't merge this

RizaneEves:
Seems like I'm not allowed to resolve conflicts but if you replace BarbarianFisherAndCooker.java with the one I have here it should be good.

ChrisBPB:
the conflicts should be fixed before you create the PR

PR #1778, opened 2017-12-18
Update v1.2 OgreCannon

iDznPB:
Added attack method, antiban, rewrote reloading and gui, updated leave rails and added walk to spot.


Strikegently:
.form files won't work on the SDN. Your UI needs to be pure Java.

iDznPB:
Forgot to remove them but all gone now.

PR #1791, opened 2017-12-30
Fixed some small bugs

gabino75:


Strikegently:
This won't show up on the SDN, because you removed the topic attribute.

PR #1792, opened 2017-12-30
Uploading fixes to ConstructUp Script

iDznPB:
Added some fixes for resizable mode to use widgets and not co-ords. Also moved a level up widget handler into main class rather than just in the butler class as those not using a butler would of been stuck.


iDznPB:
Please wait for this, I have a further commit to add.

iDznPB:
All done.

Strikegently:
I would strongly encourage you to clean this up. It would make it infinitely easier to maintain in the future. 

https://github.com/iDznPB/powerbot/blob/0ab0b6cd861d923fd9ba39af484acb5345e56d86/src/iDzn/ConstructUp/ConstructUp.java#L200-L517

PR #1800, opened 2018-01-03
Implement log. Implement statistics. Pull info from rs mediawiki.

jps333:


Strikegently:
There are potential legal issues with implementing copyrighted material in this repo that need to be reviewed prior to this being merged.

I would suggest that you use the Java standard library unless absolutely necessary.

jps333:
Oh shoot, I forgot that you had mentioned the java std library included Json parsing. I'll switch over when I get home.

PR #1808, opened 2018-01-12
WASD Camera Turning

xXTheAwesomerXx:
Adds support for using the WASD keys instead of the directional keys to change the angle and pitch of the camera.

### DIFF
--- a/src/main/java/org/powerbot/script/rt6/Camera.java
+++ b/src/main/java/org/powerbot/script/rt6/Camera.java
@@ -104,7 +104,13 @@ public class Camera extends ClientAccessor {
 			return true;
 		}
 		final boolean up = pitch() < percent;
-		ctx.input.send(up ? useWasd ? "{VK_W down}" : "{VK_UP down}" : useWasd ? "{VK_S down}" : "{VK_DOWN down}");
+		String stringToSend;
+        	if (up) {
+            		stringToSend = useWasd ? "{VK_W down}" : "{VK_UP down}";
+        	} else {
+            		stringToSend = useWasd ? "{VK_S down}" : "{VK_DOWN down}";
+        	}
+        	ctx.input.send(stringToSend);
 		for (; ; ) {
 			final int tp = pitch();
 			if (!Condition.wait(new Condition.Check() {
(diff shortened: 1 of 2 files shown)


ChrisBPB:
really the existing methods should be changed to call the new version i.e 
`public boolean pitch(final int percent) {
pitch(percent, false);
}`


to reduce duplicate code without breakages

Strikegently:
Are there any scenarios in which you can't use WASD keys?

xXTheAwesomerXx:
@Strikegently If legacy interface mode and/or the chat always on feature is enabled then wasd will be typed into the chatbox; when legacy interface mode is turned on it switches chat always on to enabled by default. The purpose of this feature into the API would be for developers who wish to have the option to use WASD (I personally use WASD when moving the camera but recently figured that the bot uses directional)

ChrisBPB:
Chat always on varpbit is (off to on)
1775 - 0 (0x0 -> 8 (0x8)

it is also the same when you turn legacy interface on, so you should use this to check if wasd is appropriate

Strikegently:
Make sure these scenarios are accounted for.

xXTheAwesomerXx:
I believe the varpbit for always on chat changes depending on legacy interface

ChrisBPB:
there are other ones that change for the legacy interface but 1775 still changes to 8 from when I tested

xXTheAwesomerXx:
[12:59:45] 1775 - 1024 (0x400) -> 1032 (0x408)
[12:59:49] 1775 - 1032 (0x408) -> 1024 (0x400)
That's what I got using legacy interface and rs3 interface

ChrisBPB:
8 still appears to be the common factor, you could probably sort it with a mask still

@ArtificialPB might know more

xXTheAwesomerXx:
Methods should work flawlessly now

Strikegently:
what is:
`if ((ctx.varpbits.varpbit(1775) >> 3 & 0x1) == 1)`

?

ChrisBPB:
check for chat always on

Strikegently:
Put this in its own method, under `Game` to improve code readability and maintainability.

xXTheAwesomerXx:
Added said method to the Game class and updated the Camera class to use that method

Strikegently:
` tx.input.send(up ? useWasd ? "{VK_W down}" : "{VK_UP down}" : useWasd ? "{VK_S down}" : "{VK_DOWN down}"); `

Could you clean these up?
Nested ternaries are incredibly cumbersome to read.

xXTheAwesomerXx:
Sorry for the delayed response code should be optimized for readability.

PR #1809, opened 2018-01-13
A drop method with shift click ability

ChrisBPB:
Reduces inventory drop time by over 15 seconds

### DIFF
--- a/src/main/java/org/powerbot/script/rt4/Inventory.java
+++ b/src/main/java/org/powerbot/script/rt4/Inventory.java
@@ -99,6 +99,21 @@ public class Inventory extends ItemQuery<Item> {
 		}
 		return ctx.widgets.widget(Constants.INVENTORY_WIDGET).component(Constants.INVENTORY_ITEMS);
 	}
+	
+	/**
+     	* Drops specified item via regular or shift dropping.
+     	*
+     	* @param i The item to drop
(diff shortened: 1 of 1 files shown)


src/main/java/org/powerbot/script/rt4/Inventory.java
```diff
@@ -99,6 +99,21 @@ public Component component() {
...
+	
+	/**
+     	* Drops specified item via regular or shift dropping.
+     	*
+     	* @param i The item to drop
+     	* @param shift Shift dropping, if true the method will verify it is enabled and fall back to regular if not
+     	* @return Success
+     	*/
+    	public boolean drop(Item i, Boolean shift){
+		if(shift && (ctx.varpbits.varpbit(1055) & 131072)>0){
```
ArtificialPB: Here's a cleaner check to see if shift dropping is enabled:
```
public boolean shiftDropping() {
        return ctx.varpbits.varpbit(1055, 17, 0x1) == 1;
    }
```

ChrisBPB:
have updated it and separated the check for potential other methods

Strikegently:
1. Can you rename `shiftDropping` so it's more apparent what it does?
Maybe `shiftDroppingEnabled` or `canUseShiftDropping`

2. Is there a reason the `shift` parameter isn't primitive?

ChrisBPB:
updated

somehow got into the habit of Boolean instead

PR #1811, opened 2018-01-15
Add worldID filter

turningwind:


Strikegently:
Rather than adding an arbitrary id filter, fit it into the appropriate architecture.

https://github.com/powerbot/powerbot/blob/rsbot-api/src/main/java/org/powerbot/script/Identifiable.java

turningwind:
@Strikegently Sorry for my ignorance. Are the new changes what you were looking for?

Strikegently:
I meant make Worlds an Id Query and override its methods to keep it polymorphic and consistent with the API.

turningwind:
@Strikegently I'm not quite sure what's wrong with the default method implementations, what functionality does there need to be added by overriding?

Strikegently:
Nothing - should be fine.

PR #1815, opened 2018-01-18
Choppenheimer

Affix:
Features :

- Task Based

 - Chops different trees

   - ENUM Based so easily extendable

- Banking
 - Single Banking class
- Walking to different spots
 - Single Walking Task to walk to Trees and Bank
- repaint
- Simple GUI to select your tree!


ChrisBPB:
looks good, only thing I can think of real quick is where you using GameObject.select...isEmpty() in your executes, #.select() will select quite a large radius of GameObjects and you may run into issues with that

I believe you can restrict your search via GameObject.select().select(int radius)... this should also reduce the intensity of the query

PR #1820, opened 2018-01-29
Viewport logic fix

kThisIsCvpv:
There is a logic error in inViewport() : boolean where the upper x and y coordinates are not bounded in resizable mode. Points that are offscreen are being registered as being within the viewport even when they are not.

Current: 
a

Fixed: 
b


Strikegently:
Neat.

PR #1823, opened 2018-02-03
Update to component

AshGreh:
The component number was changed from 33 to 32 in the latest update of the game, 33 now refers to a close button.


Strikegently:
Instead of a magic value, can you extract this to an easily identifiable constant value?

Strikegently:
I went ahead and merged this so we can get a release.

I'll clean up the API later.

PR #1829, opened 2018-02-09
hopefully fix incombat

ChrisBPB:


Timer:
This was an important check and probably shouldn't be removed, but I can't say for certain if things have changed.

PR #1830, opened 2018-02-11
AlKharid Tanner

awbasham:


Strikegently:
Don't name your main script class `Main`. The SDN identifies scripts by user ID and class name, so something more meaningful would be best.

awbasham:
@Strikegently Should be fixed now.

Strikegently:
I would also encourage you to make more meaningful commit messages.  "Quickfix" is a meaningless message that just clutters your history and makes it difficult to find a particular commit you are looking for.

awbasham:
I normally would, but they were literally "quick fixes". As in I refactored the Main.class file and it messed up a bunch of strings for some reason.

PR #1831, opened 2018-02-12
Faster valid checks for GameObjects

Strikegently:
Creates a static "NIL" instance that is lazy-loaded and used for `valid()` to improve validation speed.
99% of the time, when a script writer is calling `valid()`, they want to make sure it's not NIL. This does an additional, preliminary check that the object is not NIL.

Checking memory addresses is infinitely faster than a object query.

### DIFF
--- a/src/main/java/org/powerbot/script/rt4/Objects.java
+++ b/src/main/java/org/powerbot/script/rt4/Objects.java
@@ -14,10 +14,11 @@ import org.powerbot.script.Locatable;
  * Objects
  */
 public class Objects extends BasicQuery<GameObject> {
-	private static GameObject NIL;
+	private GameObject NIL;
 
 	public Objects(final ClientContext ctx) {
 		super(ctx);
+		NIL = new GameObject(ctx, null, GameObject.Type.UNKNOWN);
 	}
 
 	public BasicQuery<GameObject> select(final int radius) {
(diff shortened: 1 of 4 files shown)


Paris:
Would this not cause stale data or race conditions on different threads?

src/main/java/org/powerbot/script/rt4/Objects.java
```diff
@@ -14,6 +14,8 @@
  * Objects
  */
 public class Objects extends BasicQuery<GameObject> {
+	private static GameObject NIL;
```
Timer: While not currently supported, there can live multiple `Objects` with differing `ctx` so static doesn't make sense here.

Why not make this private and an instance variable, then instantiate in the constructor? That'd remove the if clause in `nil()` too.

src/main/java/org/powerbot/script/rt6/Objects.java
```diff
@@ -26,6 +26,8 @@
  * Utilities pertaining to in-game objects.
  */
 public class Objects extends MobileIdNameQuery<GameObject> {
+	private static GameObject NIL;
```
Timer: Ditto.

Strikegently:
`NIL` is no longer lazy-loaded and instead just created with the `Objects` instance per @Timer

src/main/java/org/powerbot/script/rt4/Objects.java
Strikegently: done and done.

PR #1834, opened 2018-02-24
Fix magic tab (new icon ingame)

ChrisBPB:
checked and remains the same for resizable and fixed


ChrisBPB:
also checked all the others and they have not changed

PR #1838, opened 2018-02-25
GameObjects can be reachable if target Node has object collision flags

imDeprecated:
Common sense says that a GameObject should be reachable if local player can touch it, not if local player can occupy its location


imDeprecated:
Replaces:

```
    /**
     * Determines if a game object is reachable by
     * checking the tiles around it.
     *
     * @param go The game object being tested.
     * @return True or false.
     */
    private boolean reachable(GameObject go) {
        int a = go.width();
        Tile t1 = new Tile(go.tile().x() + a, go.tile().y(), go.tile().floor());
        Tile t2 = new Tile(go.tile().x() - a, go.tile().y(), go.tile().floor());
        Tile t3 = new Tile(go.tile().x(), go.tile().y() + a, go.tile().floor());
        Tile t4 = new Tile(go.tile().x(), go.tile().y() - a, go.tile().floor());

        return (t1.matrix(ctx).reachable() || t2.matrix(ctx).reachable() || t3.matrix(ctx).reachable() || t4.matrix(ctx).reachable());
    }

```

With:

```
    /**
     * Determines if a game object is reachable 
     *
     * @param go The game object being tested.
     * @return True or false.
     */
    private boolean reachable(GameObject go) {
        return go.reachable();
    }
```

Source:  https://github.com/powerbot/powerbot/blob/scripts/src/DieselSkrt/DRuneRunner/Walker.java

ChrisBPB:
unless functionality is changed in the RSBot API this is going to break this script

Strikegently:
If we're going to be redefining terms in the API then it should be consistent across RT4 and RT6. Please make the corresponding changes to RT6.

Timer:
Reachable implies stand-on-able in the current API, so we can't change this behavior.
I'd be ok with modifying dijkstra's and adding a mode which toggles this behavior, but we have to give it a different name -- maybe `touchable()`?

imDeprecated:
Do wall lever game objects not include the wall flags?  As for usability there is wide range of usability.  As it stands right now you cannot determine if you can reach or interact with a game object through through the standard api.  All that can be asserted is that you cant occupy its location but you may be able to stand next to it (assuming there is no wall there which you also cannot test for).

As an easy example, here is a bank bot trap with the current api.  There is no way to assert that bank is not reachable without using a combination of orientation and testing the location in front of it.  This method does not work as well with deposit boxes that allow any orientation.

It seems to me that a touchable approach has far more use case and far less exceptional objects than the current approach.

image

Strikegently:
Don't close it; just make the modifications on the same branch.

Timer:
I don't see why developers couldn't specify a tile they know is reachable which allows access to said object, so like Y-1 for banks in certain orientation, etc. This greatly reduces API complexity.

Strikegently:
@imDeprecated 
Any other input?

imDeprecated:
Close for now

On Mar 7, 2018 8:00 PM, Strikegently <notifications@github.com> wrote:

@imDeprecated<https://github.com/imdeprecated>
Any other input?

Issue #1837, opened 2018-02-24
isSpellSelected


Strikegently:
Equivalent of the RT6 `client.isSpellSelected` is needed for OSRS.

Mkerian10:
Should fixed no?

ChrisBPB:
he wanted a hook for rt6 and rt4 but I am told this doesn't exist, but we have methods to achieve this already

Strikegently:
Re-opening this. The hook does exist in RT4.

Paris:
Merged into #2133.

PR #1841, opened 2018-03-01
Magic Update

kThisIsCvpv:
## Major Changes: 
- _spell()_ had it's deprecation removed.
- _spell()::Spell_ has been changed to _spell()::MagicSpell_ to cover the entire scope of magic spells instead of just the modern spellbook. **This may break scripts** that are reliant on the old return type.
- _Spell.class_ had it's _component()::int_ removed. It is pointless due to the presence of the _component(MagicSpell)::Component_ method. **This may break scripts** that are reliant on the function.

## Minor Changes: 
- Following the recent farming update, GEOMANCY, OURANIA_TELEPORT, and SPIN_FLAX have been added to the Lunar Spellbook.
- The spellbook widget now refers to an integer in the Constant class.

## Additional Comments:
- _component(MagicSpell)::Component_ is currently O(n!) where n is the size of the spell book. This can be reduced to O(n) by introducing a component variable in all the spells. This method is not very future proof friendly, which is why I refrained from doing it. The addition of any spell would shift the index of all the components by whatever it may be. However, this would fix any possible scripts that rely on previous _Spell.component()::int_ deprecated method. I can implement this if you wish.
- A method could also be added to compare the current spell with the target spell to return a boolean; answers the current open issue at https://github.com/powerbot/powerbot/issues/1837

### DIFF
--- a/src/main/java/org/powerbot/script/rt4/Constants.java
+++ b/src/main/java/org/powerbot/script/rt4/Constants.java
@@ -121,6 +121,7 @@ public final class Constants {
 	public static final int PRAYER_SELECTION = 83;
 
 	public static final int SPELLBOOK_VARPBIT = 439;
+	public static final int SPELLBOOK_WIDGET = 218;
 
 	public static final String[] BANK_NPCS = {"Banker", "Ghost banker", "Banker tutor", "Sirsal Banker", "Nardah Banker", "Gnome banker", "Fadli", "Emerald Benedict"};
 	public static final String[] BANK_CHESTS = {"Bank chest"};
--- a/src/main/java/org/powerbot/script/rt4/Magic.java
+++ b/src/main/java/org/powerbot/script/rt4/Magic.java
@@ -1,6 +1,5 @@
 package org.powerbot.script.rt4;
 
-import org.powerbot.script.rt4.Component;
 import org.powerbot.script.rt4.Game.Tab;
 
 /**
--- a/src/main/java/org/powerbot/script/rt4/Magic.java
+++ b/src/main/java/org/powerbot/script/rt4/Magic.java
@@ -34,7 +34,7 @@ public class Magic extends ClientAccessor {
 	 * 
 	 * @return The magic spell being casted.
 	 */
-	public MagicSpell spell() {
+	public MagicSpell magicspell() {
 		final Book book = book();
 
 		for (MagicSpell spell : book.spells()) {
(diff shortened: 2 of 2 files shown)


Strikegently:
Breaking changes are not allowed.

Strikegently:
The changes are fine, just make them in a way that doesn't break shit. Add a deprecated to `spell.spell()` and advise people to use `spell.MagicSpell()`

kThisIsCvpv:
## Further Changes:

### Magic.spell()
_Magic.spell()_ has been re-introduced. It is deprecated once again, the code rewritten, the comments specifying warnings and an annotation pointer to use the _Magic.magicspell()_ method.

### Spell.component()
Spell had it's constructor reverted and is back to using it's component code form. As such, _Spell.component()_ also has returned, deprecated once again, and now has an annotation pointer to use the _Magic.component(MagicSpell)_ method.

### Magic.casting(MagicSpell)
_Magic.casting(MagicSpell) :: boolean_ validates whether a spell is selected in the spell book. It is a new method to aid those who wish to make a quick validation check.

**Unit testing will follow shortly.**

kThisIsCvpv:
### magic.magicspell() - Passed
- **Standard Spell Book**
    - No Spells Selected -> Spell.NIL
    - Air Strike Selected (Have Runes) -> Spell.AIR_STRIKE
    - Air Strike Selected (No Runes) -> Spell.AIR_STRIKE
- **Ancient Spell Book**
    - No Spells Selected -> Spell.NIL
    - Ice Barrage Selected (Have Runes) -> AncientSpell.ICE_BARRAGE
    - Ice Barrage Selected (No Runes) -> AncientSpell.ICE_BARRAGE
- **Lunar Spell Book**
    - No Spells Selected -> Spell.NIL
    - Cure Other Selected (Have Runes) -> LunarSpell.CURE_OTHER
    - Cure Other Selected (No Runes) -> LunarSpell.CURE_OTHER
- **Arceuus Spell Book**
    - There's nothing here to test.

### magic.spell() - Passed

Strikegently:
@kThisIsCvpv 
Fix the cast line and this can be merged.

kThisIsCvpv:
I'm trying to validate Arceuus right now (when I've time).
I'll re-commit in a day.

kThisIsCvpv:
Keep in mind that the code still has the fault where you cannot cast spells that have the same texture across books as I described in Discord. The component ids will need to be hard-coded. Structure of the entire class may change if the code is to remain backwards compatible.

An alternate option would be checking if the component is visible but that sets the requirement that the spell book tab is open; it may become a nuisance for some scripts, but won't depend on hardcoding ids.

PR #1842, opened 2018-03-02
Added Edgeville support and a paint

Mkerian10:


Strikegently:
This won't compile. Your actual package does not match the package declaration.

Mkerian10:
I redid it. I believe that this should work.

I know that I updated all of the files, but I changed the way I represent the script in my workspace, so I think that I'll have to do this once unless I want to change the package declarations whenever I commit in the future.

Mkerian10:
Removed the script from this SDN and I'm putting it on my personal SDN.

PR #1846, opened 2018-03-03
Bugfix dismissing r event while telegrab selected

RZME:
fixes dismissing the random event while having telegrab spell selected

### DIFF
--- a/src/main/java/org/powerbot/bot/rt4/RandomEvents.java
+++ b/src/main/java/org/powerbot/bot/rt4/RandomEvents.java
@@ -34,8 +34,9 @@ public class RandomEvents extends PollingScript<ClientContext> {
 		if (!threshold.contains(this)) {
 			threshold.add(this);
 		}
-	 if(ctx.magic.spell()!= Magic.Spell.NIL){
-     	ctx.input.click(ctx.widgets.widget(320).component(Random.nextInt(1,24)).component(0).centerPoint(),true);
+	   if(ctx.magic.spell()!= Magic.Spell.NIL){
+            ctx.game.tab(Game.Tab.MAGIC);
+           
             Condition.wait(() -> ctx.magic.spell()== Magic.Spell.NIL);
         }
 		 if(ctx.inventory.selectedItemIndex() >= 0){
--- a/src/main/java/org/powerbot/bot/rt4/RandomEvents.java
+++ b/src/main/java/org/powerbot/bot/rt4/RandomEvents.java
@@ -34,7 +34,10 @@ public class RandomEvents extends PollingScript<ClientContext> {
 		if (!threshold.contains(this)) {
 			threshold.add(this);
 		}
-		
+	if(ctx.widgets.component(218, 20).borderThickness() != 0) { // if the telegrab spell is selected click somewhere in the magebook to deselect it
+            ctx.input.click(665,350,true);
+            Condition.wait(() -> ctx.widgets.component(218, 20).borderThickness() == 0);
+        }
 		 if(ctx.inventory.selectedItemIndex() >= 0){
 			ctx.inventory.selectedItem().click();
 		}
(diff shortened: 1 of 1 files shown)


ChrisBPB:
you are editing RSBot files, not your project files, so your changes must be general use - not specific to you

In order for this to be effective you should add a getSelectedSpell() or similar method to Magic (mirror changes should be added to rt6).

Additionally clicking specific co-ordinates is bad for resizable, and in general just a bad idea because you don't want 100,000 users clicking the same spot every time a random event appears.

ChrisBPB:
https://github.com/powerbot/powerbot/blob/rsbot-api/src/main/java/org/powerbot/script/rt4/Magic.java#L29


will already return the selected Spell

RZME:
tried to find the widget component for the uptext to check but couldnt find it. But regarding the action that should be taken: will ctx.menu.click(menuCommand -> menuCommand.action.equals("Cancel")); work?

ChrisBPB:
Probably not because inventory won't say cancel it would say Use or something.

At the current time, while we don't have a hook for the selected Spell, you would just for if(ctx.magic.spell()!=Spell.Nil){
//cancel the spell, but not my clicking coordinated
}

ChrisBPB:
I was saying if an inventory item is selected it would be Use not Cancel, and they may be other cases but I don't know

I imagine your way would work though so put whatever you think is best and Big Coma can review it

ChrisBPB:
formatting is messed up and probably not good idea for inline comments like that, the code itself is clean so there shouldn't be confusion from reading

also to add, on further thinking I probably would not rely on your Cancel menu thing because if the spell has runes the option will be Cast or something which still needs to be cancelled - how you handle it needs to be thought of carefully so you don't attack a random NPC, or alch a party hat - etc

Strikegently:
You can click the inventory to cancel a spell cast.

Strikegently:
What is
`ctx.input.click(ctx.widgets.widget(320).component(Random.nextInt(1,24)).component(0).centerPoint(),true);`

??

RZME:
click in a random inventory box (widget.click didnt work for me -> tested it)

ChrisBPB:
as mentioned previously you have to be careful you don't alch something that shouldn't be alched

Strikegently:
1. If the user is already on the magic tab, then `ctx.game.tab(Game.Tab.MAGIC);` won't do anything.
1. You don't do anything in the event of failure to deselect the spell.
1. Fix the formatting.

Strikegently:
1. Please fix the formatting. 
1. Use more descriptive commit messages. I'll squash your commits with a proper message, but please do it properly in the future.

src/main/java/org/powerbot/bot/rt4/RandomEvents.java
```diff
@@ -34,7 +34,20 @@ public void poll() {
 		if (!threshold.contains(this)) {
 			threshold.add(this);
 		}
-		
+	        if(ctx.magic.spell()!= Magic.Spell.NIL){
```
Strikegently: Fix the spacing.

src/main/java/org/powerbot/bot/rt4/RandomEvents.java
```diff
@@ -34,7 +34,20 @@ public void poll() {
...
+		   	} else {
+	                	ctx.game.tab(Game.Tab.INVENTORY);
+		   	}
+			      
+	           	Condition.wait(new Condition.Check() {
+				@Override
+				public boolean poll() {
+			    		return ctx.magic.spell()== Magic.Spell.NIL;
+				}
+		    	},200,10);
```
Strikegently: Spacing.

src/main/java/org/powerbot/bot/rt4/RandomEvents.java
```diff
@@ -34,7 +34,20 @@ public void poll() {
...
+	        	if(!ctx.game.tab(Game.Tab.MAGIC)) {
+		       		ctx.game.tab(Game.Tab.MAGIC);
+		   	} else {
+	                	ctx.game.tab(Game.Tab.INVENTORY);
+		   	}
+			      
+	           	Condition.wait(new Condition.Check() {
+				@Override
+				public boolean poll() {
+			    		return ctx.magic.spell()== Magic.Spell.NIL;
```
Strikegently: Spacing.

Strikegently:
@RZME 
If you fix the formatting, we can get this merged.

Strikegently:
Closing since he apparently doesn't see this.

PR #1849, opened 2018-03-15
uploading new script

samlol:
uploading my script to openSDN.


Strikegently:
This won't compile. Your package declaration does not match the actual package structure.

PR #1857, opened 2018-03-23
This was an accidental Pull request - Updated code to conform to java conventions 

Trikkstr:
(edit 1) My mistake, I was trying to update my fork. I did not mean to create this pull request on your end

(edit 2) I just pushed my newest updates onto this branch from my working branch so that this pull request could at least be useful.


Strikegently:
I'm going to say that it's not useful. You're just cluttering the history but deleting files, re-adding files, deleting files, then re-adding them again.

PR #1859, opened 2018-03-25
[RS3] Goblin Killer

SILENQER:
• Kills Goblins in Lumbridge
• Opens the door if closed
• Eats Cabbe, Cooked Meat or Cooked Chicken if HP is under 30%
• Random Movement etc. (AntiBan)


ChrisBPB:
You really should only open the door if you intend to go inside, you should also check it isn't open before trying to open it.

PR #1882, opened 2018-05-22
Reduce thread sleep

turningwind:
ClientContext has 6 PollingScripts which results in a poll every 360ms, the sleep is too long for each one so it should be reduced.


Strikegently:
@Timer / @Paris 

How about placing each of the background tasks on a separate thread?

PR #1887, opened 2018-06-04
rt4 Item boundingRect and index getters

tanlines:
boundingRect simply for being able to paint it like you would a widget.
index getter because I don't see a reason why it should be private.


Strikegently:
Rename `inventory_index` to follow Java naming conventions;

PR #1888, opened 2018-06-05
Updated goblin_killer

Trikkstr:
Added updates requested by users:
Fixed the "Fight" class so that all stackable items dropped by a kill are picked up
Only stackable items are picked up so that the bot does not have to empty its inventory
Added GUI to show runtime, exp gain, and bot status
Improved the algorithm for selecting a goblin target

There are a lot of commits because I had to remove some bits of code that I personally wanted to keep on my end


Strikegently:
Why do you have 50 commits, most of which are "do something, revert, do it again" ?

I'll squash this, but manage your history properly in the future.

PR #1893, opened 2018-06-12
Ourunia

Dragas:
Fixes an issue where if a user hadn't done Dream mentor, the script would try to interact with an empty bank.


Strikegently:
`final boolean jackIsThere = ctx.npcs.select().name("'Bird's-Eye' Jack").first().peek().valid();`
That call is incredibly inefficient. Use `!ctx.npcs.select().name("'Bird's-Eye' Jack").isEmpty()` in the future.

rs-sliske:
thats still inefficient as the value isnt really going to change from 1 poll to the next.

also how does check varp for quest progress compare to npc lookup?

Dragas:
The check is done on "start", when queue is built, not on "poll".

rs-sliske:
oh sorry wasnt paying enough attention to where the diff had split

PR #1896, opened 2018-06-29
Fixes cast glitch

4ss4s1n:
Cast glitch is where the mouse keeps hovering over npc when a spell is selected.
Response to:
https://www.powerbot.org/community/topic/1346112-random-event-issue/?tab=comments#comment-16597202

PR #1899, opened 2018-06-29
Fixed spacing

4ss4s1n:
Follow up of https://github.com/powerbot/powerbot/pull/1896


4ss4s1n:
I also think the logic can be applied inside the loop, start on [L44](https://github.com/powerbot/powerbot/blob/9a93564dab0889da8415193fd4b6126d8d1ded66/src/main/java/org/powerbot/bot/rt4/RandomEvents.java#L44).

Strikegently:
You don't need to close a Pull Request when changes are requested. Push to your branch and the PR will update accordingly.

4ss4s1n:
@Strikegently thoughts on applying the logic inside the loop?

4ss4s1n:
@Strikegently thoughts on applying the logic inside the loop?

Strikegently:
You seem to be over-complicating it. 
If an item is selected, deselect it.
If a spell is selected, deselect it.

That's really all you need.

4ss4s1n:
I personally think this is more readable. You shouldn't have to make a check for item being selected when you're not in the inventory. Yes, right now it's easy to have 2 cases, but in the future, as osrs keeps updating you may have trouble keeping up with all the cases. Also, this allows you to add more to each tab case specifically. I can roll it back to just checking the spell if you want, your call.

Mkerian10:
@4ss4s1n But regardless you'll have to continue making cases in the future. Plus it's been like this for years, odds are it won't change significantly in the future.

PR #1904, opened 2018-07-12
Updating Bank#withdraw()

Mkerian10:
Currently banking can take a while since it validates that it successfully withdrew the item. I'd like to allow the Script Writer to be able to choose to validate it himself, and instead move on to the next item to withdraw. This will help a lot for bank standing scripts as banking currently takes much too long in it's current form.

Even if we don't want to implement this I would definitely appreciate a future update bringing an option to withdraw without the wait.

### DIFF
--- a/src/main/java/org/powerbot/script/rt4/Bank.java
+++ b/src/main/java/org/powerbot/script/rt4/Bank.java
@@ -246,7 +246,10 @@ public class Bank extends ItemQuery<Item> {
 		} else {
 			action = "Withdraw-X";
 		}
-		final int cache = ctx.inventory.select().count(true);
+		
+		final int cache;
+		if(!validate) cache = ctx.inventory.select().count(true);
+		
 		if(!item.component().visible()){
          	   ctx.bank.currentTab(0);
         	}
--- a/src/main/java/org/powerbot/script/rt4/Bank.java
+++ b/src/main/java/org/powerbot/script/rt4/Bank.java
@@ -213,9 +213,10 @@ public class Bank extends ItemQuery<Item> {
 	 *
 	 * @param item   the item instance
 	 * @param amount the amount to withdraw
+	 * @param validate whether to check if the item was successfully withdrawn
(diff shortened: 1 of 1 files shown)


Strikegently:
I don't believe checking the contents of the inventory is that expensive of a call. I'm almost certain the bottleneck would be elsewhere.

Mkerian10:
@Strikegently It's not that it's expensive, but it's the Condition.wait() aspect of it. It doesn't end the function until the contents of the inventory has updated, this adds about a half second wait after every time, which is un-humanlike and annoying for stuff where you need speed. I can send a video if you want.

ChrisBPB:
if I remember I will add some debugging to withdrawing later today on a local build and see if it is the issue

ChrisBPB:
Ok so in general either the first time you open the bank, or the first time the scripts ran (I'm not really sure at this point) the first withdrawal can take up to 500ms to create the cache (final int cache..) following calls will take around 150ms.

2nd thing that seems to take between 700 and 1100 ms is this whole if statement
https://github.com/powerbot/powerbot/blob/rsbot-api/src/main/java/org/powerbot/script/rt4/Bank.java#L252

finally the condition wait takes between 500 and 800 but completes on its first loop so again the bottleneck is the .count(true) call not the fact that it's looping until it's condition is met

you'd only be saving a little bit of time by ignoring the condition but I do kind of agree a human would withdraw as fast as possible and would use the fact they clicked to confirm the withdrawal rather than waiting to visually see and count the items in the inventory

**anyway, with all that said this implementation unless I missed something seems to break functionality in that it oly really considers Withdraw-X, and will return false (https://github.com/Mkerian10/powerbot/blob/a3c83e683f8be894e6c916c8d6b7fcf838d2e049/src/main/java/org/powerbot/script/rt4/Bank.java#L281) no matter what if you - this includes breaking Withdraw-X functionality where it has already been called once (because it will now already be on the menu as Withdraw-14 for example).**

Strikegently:
In the near future, a more holistic refinement of the banking API will be undertaken with many of your concerns addressed.

PR #1905, opened 2018-07-12
Update bank to add escape closing

Mkerian10:
Adds a method for closing the bank with escape, also checking for escape closing to be active.


Strikegently:
Shouldn't it wait for the bank to close?

Mkerian10:
@Strikegently It's been updated, that was a silly mistake by me. I'm not sure if the API doesn't allow lambdas for some reason because most of the code is in anon form, just let me know and I can change it. Now it should properly check for a closed bank, and then close it with keys if it isn't. My only question is, right now the way I have it structured, if the Condition.wait() fails, it'll go to the default close() method. Would you rather just return false if the escape fails or have it manually close?

Strikegently:
This won't compile.

Mkerian10:
@Strikegently Why not? It's working fine on my machine.

Mkerian10:
@Strikegently Should be fixed, dunce mistake

Issue #1886, opened 2018-06-04
[OSRS] Request for methods to get Item boundingRect and index, boundingRects/points for inventory slots


tanlines:
Some activities are done most efficiently with a specific inventory setup which requires dragging items to specific inventory indexes.

ChrisBPB:
Unsure of what you are asking. Items already have #boundingRect() and #inventoryIndex() methods

ChrisBPB:
oh for the slots themselves?

Mkerian10:
@ChrisBPB He wants the bounding rects for the inventory slots.

ChrisBPB:
They used to be widgets but I can't find them anymore, any idea on this @Strikegently ?

Mkerian10:
https://github.com/powerbot/powerbot/pull/1906

PR #1906, opened 2018-07-12
Adds dragging and indices to RT4.Inventory

Mkerian10:
Adds functionality for finding points at various indices, also adds a function to get the boundingRectangle as well as offers dragging functionality for dragging an item to a certain index.

The boundingRectangle/point is off slightly during banking. This is because the default Inventory#component() changes for some reason, I'm not sure why. Widget 140 remains correct even during banking however it may be something I don't understand.

This resolves most of Issue #1886, and there's another commit that I believe has already been merged would resolve the rest of it. Which would end Issue #1886.

### DIFF
--- a/src/main/java/org/powerbot/script/rt4/Inventory.java
+++ b/src/main/java/org/powerbot/script/rt4/Inventory.java
@@ -142,6 +142,73 @@ public class Inventory extends ItemQuery<Item> {
 			return true;
 		}
 	}
+	
+	/**
+	 * Finds the centerPoint of the inventory's index
+	 * @param index 0-27, index of inventory
+	 * @return centerPoint of the index param
+	 * @throws IndexOutOfBoundsException if index is below 0 or above 27
+	 */
+	public Point indexCenterPoint(int index){
+		if(index < 0 || index > 27){
+			throw new IndexOutOfBoundsException();
+		}
+		
+		final Point base = component().screenPoint();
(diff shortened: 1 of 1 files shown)


Strikegently:
Are coordinates affected by fixed/resizable screens?

Mkerian10:
@Strikegently Coordinates are correct for both. Only time they're not from what I can see is during banking.

https://imgur.com/a/D1u14M6

AshGreh:
Could we have from index to index instead of a cached Item if this is accepted @Strikegently

Mkerian10:
@AshGreh that’s a good idea, I’ll add another commit with that

Mkerian10:
On second thought I don’t think that’s ideal. I think we should have a method to get the item at a specific index. But overall I can’t think of a scenario where I’d need to drag anything other than an item.

I think requiring an item will reduce the amount of mistakes made by bad programmers since if the my make a mistake and there’s no item at the index, it’ll return false since it’s not valid rather than switching nothing.

I think what’d be best is adding an itemAt() for Inventory. And then maybe adding an overloaded method that accepts 2 items. Maybe a private method that moves through 2 indices, but I don’t see why it’d need to be public. I think new programmers may misuse it.

Mkerian10:
@Strikegently Is this fine or do you want other stuff added? In general I think this is fine for most purposes.

ChrisBPB:
please be sure to add imports in future as this would not compile

PR #1911, opened 2018-07-22
Adds escape closing for the bank.

AshGreh:
Updates closing by using escape key.
Has better Human Emulation than the currently open PR and actually compiles.
In case escape closing is not enabled in settings, sends a warning in the console stating that it is using the alternate procedure due to not having the ability to escape close.


Mkerian10:
No need to write a warning to the log in the event of escape closing not being enabled. It's unneeded.

Why hold escape down until the bank closes, absolutely horrible human emulation.

AshGreh:
I have done my research, most humans hold the escape button till the bank is closed, but I guess people who use buzz words generally don't know how to do research. The warning is for script developers who don't know what the problem is while debugging a script run by the user. Please don't leave comments that don't contribute to the given.

PR #1913, opened 2018-07-23
Area tiles public

tanlines:
Making the tiles public so we can see exactly what tiles are included.


Strikegently:
1. Keep `getTiles()` private
2. Add a public `tiles()` method that returns `tiles.clone()`

tanlines:
Correct me if I'm wrong.
In getTiles(), it uses Arrays.copyOf() because it doesn't matter that it's a shallow copy since the local array is only accessed once.
For a method to return tiles, you don't want the user modifying the tiles in the area, so you make a clone which is a deep copy.

PR #1937, opened 2018-09-03
remove animation check for default break

ChrisBPB:
I believe majority of scripts are relying on the default #canBreak() thus it becomes a very bot-like thing when scripts only ever log out or pause after you have finished mining, or after you've finished fletching etc.

Since the new break handler will support actual breaks not just stops, it is much better if a user can pause mid-action. (i.e go checkout facebook while their character continues to cut a tree down).

Combat checks still remain, and a break will not engage during combat. However, if you are close to aggressive monsters you should be overriding canBreak yourself!


ChrisBPB:
Going to leave this open for a day or so to see if there is any objections but it seems standard.

Strikegently:
It should be fine as long as we're not in combat.

PR #1951, opened 2018-09-29
Adds methods to Bank and Inventory.

AshGreh:
Adds the methods withdrawModeQuantity( for getting the Amount that is selected for withdraw mode) and quantityComponentValue ( for getting the component value of a quantity ) to Bank and adds Constants for them.

Adds the method isFull() to Inventory, updates the Constants for this changes and makes use of the Constant in Inventory.

### DIFF
--- a/src/main/java/org/powerbot/script/rt4/Inventory.java
+++ b/src/main/java/org/powerbot/script/rt4/Inventory.java
@@ -159,7 +159,7 @@ public class Inventory extends ItemQuery<Item> {
 	 * @throws IndexOutOfBoundsException if index is below 0 or above (Constants.INVENTORY_SIZE-1)
 	 */
 	public Point indexCenterPoint(int index){
-		if(index < 0 || index > Constants.INVENTORY_SIZE){
+		if(index < 0 || index > Constants.INVENTORY_SIZE - 1){
 			throw new IndexOutOfBoundsException();
 		}
 		
@@ -224,7 +224,7 @@ public class Inventory extends ItemQuery<Item> {
 	}
 	
 	public boolean isFull() {
-		return ctx.inventory.select().size() > Constants.INVENTORY_SIZE-1;
+		return ctx.inventory.select().size() >= Constants.INVENTORY_SIZE;
 	}
 
 	@Override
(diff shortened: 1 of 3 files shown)


src/main/java/org/powerbot/script/rt4/Inventory.java
```diff
@@ -154,12 +154,12 @@ public boolean drop(ItemQuery<Item> items) {
...
 	 * Finds the centerPoint of the inventory's index
-	 * @param index 0-27, index of inventory
+	 * @param index 0-(Constants.INVENTORY_SIZE-1), index of inventory
 	 * @return centerPoint of the index param
-	 * @throws IndexOutOfBoundsException if index is below 0 or above 27
+	 * @throws IndexOutOfBoundsException if index is below 0 or above (Constants.INVENTORY_SIZE-1)
 	 */
 	public Point indexCenterPoint(int index){
-		if(index < 0 || index > 27){
+		if(index < 0 || index > Constants.INVENTORY_SIZE){
```
ArtificialPB: Should be `index > Constants.INVENTORY_SIZE - 1` since you're dealing with an index.

src/main/java/org/powerbot/script/rt4/Inventory.java
```diff
@@ -219,10 +219,13 @@ public boolean drag(final Item item, final int index){
 		return ctx.input.drag(Calculations.nextPoint(r, objectRectangle), true);
 	}
 
-
 	public boolean shiftDroppingEnabled() {
 		return ctx.varpbits.varpbit(1055, 17, 0x1) == 1;
 	}
+	
+	public boolean isFull() {
+		return ctx.inventory.select().size() > Constants.INVENTORY_SIZE-1;
```
ArtificialPB: I would suggest you to improve readability by changing it to `ctx.inventory.select().size() >= Constants.INVENTORY_SIZE`

Strikegently:
`return ctx.inventory.select().size() >= Constants.INVENTORY_SIZE;`
Why wouldn't you just do `==` ?

AshGreh:
@Strikegently It's just a style preference. Could just go with equal to.

PR #1956, opened 2018-10-03
Fixes dragging

AshGreh:
Currently, dragging won't work for fixed mode users. This enables that, with addition of methods centerPoint( to get center point of an inventory spot (x, y) ) and indexPosition ( to get inventory position (x, y) of an inventory spot index ).


ArtificialPB:
Is the addition of these two methods really necessary? Can you give me a use-case example that's not possible/difficult with just indices?

AshGreh:
@ArtificialPB it makes everything easier when you are dealing with pattern dropping. You're able to directly work with 2D positions. It's just there for readability and also present to avoid redundant code that a few script writers have to write. I could change it if it's a problem, but in general, this is good to have.

ArtificialPB:
Alright, I would just suggest you to replace the `Point` argument with `int x, int y` in `centerPoint` method. Or just overload it. In my opinion it would be clearer what it does and would be easier to use, especially in a loop.

AshGreh:
@ArtificialPB I initially wanted to return a Pair for indexPosition instead of the current choice of point. I'm not sure at the moment if powerbot allows the package. How would you go about it? I considered int[] too for a minute, but that potentially is worse than Point.

ArtificialPB:
IIRC the security manager or the SDN blocks the javafx package so the Pair would be a no-go. 

How about splitting the function into two: `inventoryColumn(index)` and `inventoryRow(index)`. What I would also suggest you to do is, instead of having functions returning information about an item (center point, bounds), return the actual item so the users have more options (overloading `itemAt(row, column)`). 

And while you're at it, you can optimize the code at [this line](https://github.com/powerbot/powerbot/blob/8a9654d58eddcbea0e063c7677d3b50f144498a5/src/main/java/org/powerbot/script/rt4/Inventory.java#L83) so it will just get the item we want instead of loading the whole inventory :)

AshGreh:
@ArtificialPB the main reason I want to retain center point and bounds is that you cannot get bounds and center point till the item exists. What are your thoughts on this?

ArtificialPB:
I see this is needed for dragging items in the inventory, never mind then.

The only thing I see left that should be improved is the `itemAt(index)` function. Right now it only works if one of the inventory replacements is valid, but not for the plain inventory. Items in inventory component are stored in arrays of ids and stacks: https://github.com/powerbot/powerbot/blob/rsbot-api/src/main/java/org/powerbot/script/rt4/Inventory.java#L70

AshGreh:
@ArtificialPB Ah, I thought everything was wrapped in the component call, don't know why it isn't. Any who, without changing something that could break the whole of Inventory, here is my latest PR. x D

Strikegently:
Most of the methods you added should be private. They are generally helper methods which shouldn't be public.

AshGreh:
@strikegently they are exposed because a good portion of anti patterns make use of something similar. Might as well avoid redundancy.

Strikegently:
No breaking changes are allowed. You renamed a method, which would break any script that used it.

AshGreh:
@Strikegently if you could specify which one, that'd be great

ChrisBPB:
I imagine indexCenterPoint but I could be more

AshGreh:
I will fix it tomorrow, until and unless someone else wants to do the honours. x D

AshGreh:
@Strikegently you tell me https://github.com/powerbot/powerbot/blob/f16dd105ff99d76c63d735994c32a1c560610b4b/src/main/java/org/powerbot/script/rt4/Inventory.java#L31

ArtificialPB:
`6512` seems to be an inventory spot placeholder or something similar.

AshGreh:
@ArtificialPB what's 16512 for? same thing? also, would they ever be encountered in your opinion?

ArtificialPB:
No idea, I was just guessing for `6512` because it seemed like a logical conclusion based on the information we have about it.

Strikegently:
One more thing: format your code.
In IntelliJ: Ctrl + Alt + Shift + L. Select "Only VCS Changed Text"

AshGreh:
@Strikegently is the formatting currently broken?

Strikegently:
@AshGreh 
You have lines like: `return row*4 + column;`
The spacing needs to be consistent.

AshGreh:
@Strikegently didn't get a notification, will solve in a few, could you please reopen this.

Strikegently:
I squashed and merged.

Issue #1910, opened 2018-07-19
Add a Thread#isInterrupted() check to Input#apply()


Mkerian10:
I was thinking it'd make prediction and fluid gameplay much easier if we could use an Executor service to send inputs, and be able to cancel out in the middle of an input without penalty. I think it'd help out with animation cancelling and tasks that require constant input in a system where there's not always reliable immediate feedback.

It'd add a more human-like dimension in my opinion,  while also allowing us to do more complex activities even better.

Mkerian10:
Merged.

PR #1960, opened 2018-10-10
Adding Thread#interrupted check to apply

Mkerian10:
I'm not sure if this is the best way to do this, however I wanted the ability to cancel out of anything by interrupting the thread and this seemed most effective.

Solves #1910


ChrisBPB:
how is this going for you in the test build?

Mkerian10:
@ChrisBPB Good, I think. I don't have an tests or numbers about how it's working so honestly you could've forgotten to add it and I'd probably not notice the difference, but it does seem like it's helping out a ton. Generally I'm using it when I need to switch prayers very quickly and it seems to be working well then. I'd definitely like for it to be added in the next update, as my Zulrah script would be much worse without it.

PR #1973, opened 2018-11-04
Update Constants.java

anwar6953:
`ctx.lobby.tab()` currently returns `Tab.NONE`, which makes a lot of other functionality not work, either. For instance, `Lobby#worlds` also returns an empty list because of this.


ChrisBPB:
perfect thanks

PR #1982, opened 2018-11-11
Remove code duplication

anwar6953:


Strikegently:
Try this instead:
1. Delete the Constants.BANK_LOAD1 - BANK_LOAD10
2. In the method, check that `set >= 1 && set <= 10`. If not, return false.
3. Delete the entire switch-case block and replace it with `return bankPresetContainer.component(set).click();`

PR #1995, opened 2018-11-20
Update `Bank#presetGear` to only use shortcut for first two sets.

anwar6953:
The shortcut for sets 3 and onwards is not just a simple number press.


ChrisBPB:
set==1 or 2 should probably come first in the statement as we can avoid doing the other checks if one of those is false, and it's most likely to be the thing that is false within that statement

anwar6953:
I will try
```
ctx.input.send("{VK_CONTROL down}");
ctx.input.send(key);
ctx.input.send("{VK_CONTROL up}");
```
For keys 3+.

PR #1996, opened 2018-11-20
Update the widget used for summoning interactions.

anwar6953:
We can reuse the combat bar widget, as the summoning component is a component of that.


ChrisBPB:
Isn't the combat bar a toggle on/off? meaning not everything will have that widget on their screen?

anwar6953:
The action bar is toggle-able, but the combat bar is guaranteed to be there (except on legacy mode).
Do we support legacy mode rs3?

PR #2003, opened 2018-11-27
added needed method from inheritance

SCP-914:
Honestly don't know what this method does but it's needed by inheritance.


Mkerian10:
No it doesn't, should work fine without it

SCP-914:
> No it doesn't, should work fine without it

ok, would throw a compile error for me

ChrisBPB:
It's your script, we can merge it if you want.. but it's not needed in any way.

PR #2004, opened 2018-11-27
Updated widgets for better use of UI

SCP-914:
Now clicks the range and selects the first item to cook to prevent eating any fish.


SCP-914:
changing some code to fix misclicks

PR #2007, opened 2018-11-27
changed some sleep timers

SCP-914:
changed some sleep timers to seem less bot like


ChrisBPB:
You'd probably be better using the Conditional Waits.

PR #2011, opened 2018-11-29
More sensible mouse speed methods

AshGreh:


AshGreh:
@Strikegently it would be nice to tell me why this was closed

AshGreh:
@Strikegently this is the part where you tell me what the problem is that made you close this in the first place.

benpoulson:
I'm guessing that he's not happy with deprecating a very commonly used method, without either community support, obvious gain, or an undisputable reason.

I do think the speed methods are very confusing, however I think we need to come to a joint decision from the community and maintainers, not just the opinion of yourself. :)

AshGreh:
@benpoulson there is a quite good reason to deprecate it. Most people don't understand the usage of the method and hence, it would be better to push people towards an understandable method. As for deprecation meaning here, powerbot doesn't actually remove methods( from my time here ). Deprecate is used here to keep the method to avoid breaking older scripts while indicating to the new developers to avoid it.

Agreed, which is why I brought this up in the first place. I have had quite a good number of people asking me why it's not in percentage. Even though STS and your suggestion of 0-1 makes sense to me, it won't to others. It's better to follow percentage threshold over everything else from those observations.

AshGreh:
@Strikegently could I have your thoughts on this PR?

Mkerian10:
@AshGreh You have a duplicate method so this won't compile. Also an exception should be thrown if the given speed is out of bounds

AshGreh:
I personally think if a person is dealing with a method as sensitive as this and can't read the documents properly for the bounds, it's better to avoid informing them. It's pointless to enforce a catch block and make everyone else's code ugly just for a select few.

Mkerian10:
@AshGreh Should be an unchecked exception

AshGreh:
Even though I personally do not agree with the exception, I have added it,

AshGreh:
It has been a year since Coma's last response.

Strikegently:
it shouldn't throw a generic runtime exception if the argument is out of bounds. There's a reason [`IllegalArgumentException`](https://stackoverflow.com/questions/15208544/when-should-an-illegalargumentexception-be-thrown) exists.

Mkerian10:
@Strikegently I tried to tell him

AshGreh:
@Strikegently done.
@Mkerian10, last I checked, your name was Showtime on the discord channel, not Nox.

Strikegently:
Are you sure your calculations are correct?

AshGreh:
Rechecked it just now.
Translations are valid as per VBA.
Test 0: Value=0 SpeedInParisUnits=100.
Test 1: Value=100 SpeedInParisUnits=10.
Test 2: Value=49 SpeedInParisUnits=56.

Strikegently:
I don't want two methods for setting mouse speed that behave completely differently. We have a standard and it will be kept.

AshGreh:
@Strikegently oh is that right? You couldn't say this 3 months ago, could you? You had to make me wait and modify the code to reject it. Good job on keeping the OpenSDN closed to contributions even when your standards are the most idiotic.

PR #2014, opened 2018-11-30
More sensible mouse speed

AshGreh:
#2011 was closed without any suggestions or reasons behind it.
A good portion of the people don't seem to understand why mouse speed should be set in the way paris sets it. By using the formula 100 - (s * 0.9), you can get a percentage of the maximum speed the mouse can travel at.

### DIFF
--- a/src/main/java/org/powerbot/script/Input.java
+++ b/src/main/java/org/powerbot/script/Input.java
@@ -55,12 +55,12 @@ public abstract class Input {
 	
 	/**
 	 * Returns the current speed.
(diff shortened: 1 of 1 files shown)


Mkerian10:
Would be better to take a double and scale (.1, 1]

AshGreh:
@Mkerian10 my idea seems more sensible from the conversations I had. Most people assume that it would be a scale of 1-100 and it would be closer to what paris had in my mind. Not a bad suggestion, might add something to the method to accept it.

Strikegently:
Don't open a new PR if one was closed. If you want to discuss a PR being reopened, then tag someone.

PR #2020, opened 2018-12-15
Updating the widget for bank tabs

iDznPB:
The folder for all the tabs was changed to 11, updating now allows for tab changing to work.


Strikegently:
this won't merge because it has merge conflicts.

ChrisBPB:
I've committed this

Issue #1943, opened 2018-09-19
Allow Polygon Bounds


AshGreh:
Better tools for mapping clicks and polygon bounds would be nice to have.

ChrisBPB:
Please add more detail to your request as it is unclear what you want.

AshGreh:
Ah, sorry. I wanted to allow the ability to add a polygon/set of polygons instead of a cuboid for well, most if not all interactable entities. Also, a tool that would allow easy working with polygons to shape bounds.

ChrisBPB:
The issue that jumps out at me mainly would be NPC's. Given they are animated their model changes which is why we use a cube to surround them completely but also their orientation which may break bounds harsher than if they were a simple cuboid. 

We could make bounds adapt with the orientation of the entity but I'm not sure what cost that may have on resources, and it would not help still with the animation.

I imagine for game objects polygonal bounds wouldn't be an issue but could you give the use case you had in mind as in my head I picture most objects in RS to be quite cube'y in shape anyway

AshGreh:
I agree, it would be a bit harsher. However, it would be easier to attach bounds from cache models according to another developer. We were working on attaching cache bounds/get bounds at RT, however, we had to reflect most of it for attachment/overriding a few things that wouldn't be possible on the SDN level.

In general, it would be costly, but would give refined clicks. Maybe as an optional refinement.

For example, redwood trees. Having a cuboid would make it hard to click on a few parts of the tree. Either the developer would have to provide extra spacing( induced miss clicks ) or reduce spacing ( bound limiting ). Both in general, even though work, would eventually boil down to being a pattern.

Mkerian10:
Reading models from the cache would be nice. Would allow users to add that to their paints and also allow for more accurate clicking. I can't imagine there'd be much more complex calculations.

AshGreh:
Found another thing today. Half of the tree is empty when you specify it with a cuboid. It's horrible in my opinion that there is a good chance that the client would click on the empty area of the tree.

ChrisBPB:
Am looking into this just not sure of the best implementation right now so evaluating options.

Mkerian10:
@ChrisBPB wym best implementation? Seems like just adding the model bounds to the CacheConfigs and adding the back end to get the models from the cache would be sufficient. Design is straight forward here isn't it?

ChrisBPB:
[...]

PR #2021, opened 2018-12-18
Allow extending of specific classes

benpoulson:
Currently these four classes are unable to be extended from.
I have some interesting features I would like to potentially propose in the near future, so having these classes open would definitely aid development.


ChrisBPB:
I believe you will find issue with the BasicObject parameter as a BasicObject will require ReflectProxy which required Reflector, or am I overlooking something?

benpoulson:
That makes perfect sense. If I'm to be honest with my intentions; this was an indirect attempt to gain access to the `boundingModel` field in `org.powerbot.script.rt4.Interactive` through the parent classes.

I'm working on polygon bounding boxes ( see issue #1943 ) and I've subclassed `BoundingModel` to create `PolygonBoundingModel`; however there doesn't seem to be any direct access to the aforementioned `boundingModel` field, so I'm unable to actually use my polygon class.

As a counter proposal, would you accept a MR for the `Interactive` class which would allow me to directly get and replace the current bounding model?

ChrisBPB:
I personally don't see any issue with allowing the model to be set/get

PR #2023, opened 2018-12-20
Implement getters and setters for Interactive bounding boxes

benpoulson:
Leading on from the conversation in https://github.com/powerbot/powerbot/pull/2021, I'm requesting getters and setters for the `boundingModel` on `Interactive` on rt4 and rt6.


anwar6953:
Thanks for keeping the two APIs consistent.

PR #2025, opened 2019-01-07
Deprecate inCombat refactor into a new method

Mkerian10:
inCombat was very confusing for new people as its name is unintuitive unless you read the documentation. Deprecating it will discourage its use, and adding the new healthBarVisible method will still allow people to use the functionality when needed

### DIFF
--- a/src/main/java/org/powerbot/script/rt4/Actor.java
+++ b/src/main/java/org/powerbot/script/rt4/Actor.java
@@ -110,12 +110,7 @@ public abstract class Actor extends Interactive implements InteractiveEntity, Na
 	 */
 	@Deprecated
 	public boolean inCombat() {
-		final Client client = ctx.client();
-		if (client == null) {
-			return false;
-		}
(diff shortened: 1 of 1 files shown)


AshGreh:
how about making inCombat rely on interacting entity instead of deprecating it?

Mkerian10:
@AshGreh That would potentially break scripts relying on the current functionality. I doubt there are any that do that, and if they do they should probably be removed anyway.

I wouldn't be opposed to it though, the problem is it could be misleading as well as afaik there's no way to check that the type of interaction is attacking as opposed to following or something similar. But I doubt Coma would allow it.

Mkerian10:
Thinking about it further inCombat shouldn't be based on interaction. It obfuscates the underlying functionality since it would get triggered on things like NPCs following or talking to you.

AshGreh:
After further talk, it seems that this is the right way to move forward. Too many scenarios would not be covered with inCombat and hence, it's best to delegate the responsibility to the SW in my opinion too.

PR #2028, opened 2019-01-22
Implemented rt4 and rt6 component queries.

ArtificialPB:


Strikegently:
I like the additional filters you've added, but take a look at the latest version of the Component Query. I modified how it loads the component tree a while back that significantly improved performance, as well as added several overloads that you're missing. Since a considerable number of people are using mine, I'd like to keep it standardized for now.

ArtificialPB:
As far as I can tell, component tree loading is already optimized as per your gist, but I've "extended" it by providing an option to disable child loading and changed single widget id param to varargs.

I'll add the remaining filter overloads that are missing.

Strikegently:
One more change: add the `components` field to the context classes.

PR #2030, opened 2019-01-25
Fixing prayer components again after update.

iDznPB:
I somehow added the wrong components for some of the low level prayers.


ChrisBPB:
thanks

PR #2033, opened 2019-01-27
Fix viewport for world list scrolling

tanlines:
Fixing world list scrolling by adding a new method to get the viewport.

I didn't realise the world list was also being used as the scroll viewport. The new list is now > 2000 pixels high and doesn't work for scrolling.

### DIFF
--- a/src/main/java/org/powerbot/script/rt4/World.java
+++ b/src/main/java/org/powerbot/script/rt4/World.java
@@ -187,7 +187,7 @@ public class World extends ClientAccessor
 			if (c.index() % 6 != 2 || !c.text().equalsIgnoreCase("" + number)) {
 				continue;
 			}
-			ctx.widgets.scroll(c, list, bar(), true);
+			ctx.widgets.scroll(c, viewport(), bar(), true);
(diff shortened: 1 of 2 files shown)


ChrisBPB:
I'm not sure if viewport is a good name, viewport really would be the games viewport and could be confusing in this context, but will let @Strikegently review it

Strikegently:
1. Rename it to `container` so it's clear that it refers to the container component that has all of the world widgets;
2. Is relying exclusively on the width/height of the component smart? 
3. This would crash any script using it if point (2) doesn't check out because it returns null.

['ComponentQuery`](https://github.com/powerbot/powerbot/pull/2028) was added to the API recently. Try to find distinguishing properties to dynamically find the necessary components, using the query.

tanlines:
The component which I called "viewport" does not have any child components, so calling it container doesn't make sense. I called it viewport because that's what Widget#scroll calls it. https://github.com/powerbot/powerbot/blob/rsbot-api/src/main/java/org/powerbot/script/rt4/Widgets.java#L101

Do you want me to rename list to container?

PR #2035, opened 2019-01-27
Update prayer Effect

tanlines:
Updated correct prayer indexes.
Refactored to use quickSelectionIndex rather than ordinal() which is highly unintuitive and inflexible.

Also tested cases where effect.quickSelectIndex() was used for setting up quick prayers and seems good.

In the quick prayer setup interface, the component indexes of the prayers match the varpbit offset to check if the prayer is active.
In the normal prayer interface, the component indexes follow a different ordering.

### DIFF
--- a/src/main/java/org/powerbot/script/rt4/Prayer.java
+++ b/src/main/java/org/powerbot/script/rt4/Prayer.java
@@ -159,43 +159,45 @@ public class Prayer extends ClientAccessor {
 	}
 
 
+	/**
+	 * Order of effects is based off the quick prayer setup component ordering.
+	 * Indexes are based off the main prayer interface components.
+	 */
 	public enum Effect {
-		THICK_SKIN(5, 0, 1),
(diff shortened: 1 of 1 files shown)


ChrisBPB:
Looks ok to me but will wait for @Strikegently

Strikegently:
>Refactored to use quickSelectionIndex rather than ordinal() which is highly unintuitive and inflexible.
This seems strange to me. All you did was add the ordinal index to the constructor of each of the Prayers. You're using the exact same information, just adding duplicate entries for it. 

Think of this scenario:
A new Prayer is added. It sits between "Smite" and "Sharp Eye." You now have to update the `quickSelectionIndex` for every Prayer after Smite. Using `ordinal()`, you just add the entry and it's done.

You're adding additional complexity and burdensome maintenance for no reason.

ChrisBPB:
thanks

PR #2038, opened 2019-02-03
Lofty scripts

en-lofty:
Added Clay Humidifier Beta


ChrisBPB:
Please remove the .gitignore file

ChrisBPB:
sorry for the confusion, I should have said to remove the changes from the main

https://github.com/powerbot/powerbot/blob/scripts/.gitignore

en-lofty:
Alrighty made the changes. Is everything ok now?

ChrisBPB:
your file still has changes, they are meaningless but there is no point letting it go because you may get a similar thing in the future and resolving this may give you the knowledge on how to do it in the future also

you should be able to do the following:
git checkout HEAD^ /.gitignore

This will reset the specific file to what it was, then you will need to commit your changes and push as normal


If you're using the GUI I am really familiar with that for this scenario

en-lofty:
"they are meaningless but there is no point letting it go because you may get a similar thing in the future and resolving this may give you the knowledge on how to do it in the future also"
I agree!
I am using Intellij along with the command-line. Thanks for the help!

ChrisBPB:
perfect, for future reference you can also do HEAD^2 and HEAD^3 etc (HEAD^ is HEAD^1 effectively), 1 being the current head, 2 being the one before current etc

en-lofty:
Thanks for the tip!

Issue #2048, opened 2019-04-18
Make CacheWorker public


tanlines:
https://github.com/powerbot/powerbot/blob/rsbot-api/src/main/java/org/powerbot/script/Bot.java#L48

It says it's public in the repo but it's not in the actual jar.

PR #2049, opened 2019-04-18
Make CacheConfigs#load public

tanlines:
Requires CacheWorker to be made public to work, issue here https://github.com/powerbot/powerbot/issues/2048


ChrisBPB:
if there any issues with these methods being public @Strikegently

PR #2052, opened 2019-05-08
Fixes indexCenterPoint

AshGreh:
This will fix indexCenterPoint and methods that depend on it, such as drag.


AshGreh:
The current indexCenterPoint function
beforePatch
My patch
afterPatch

ChrisBPB:
perfect thank you

PR #2057, opened 2019-05-17
Refactor rt4 Combat to use component query

tanlines:
Use special attack orb as backup for special attack

### DIFF
--- a/src/main/java/org/powerbot/script/rt4/Combat.java
+++ b/src/main/java/org/powerbot/script/rt4/Combat.java
@@ -80,10 +80,15 @@ public class Combat extends ClientAccessor {
 		if (specialAttack() == select) {
 			return true;
 		}
+		final Component c;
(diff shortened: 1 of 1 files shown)


ChrisBPB:
logically this seems fine but I don't currently have a members account to test, did you run this to test?

tanlines:
I have tested and it works, but since changing to the Attack tab almost never fails, it basically hasn't changed behaviour.

I also realised I missed an edge case where the weapon can't use special. Unlike with the special attack bar where the component disappears if it's not there, failing the visible check, the special orb needs to check texture.

tanlines:
Case no special weapon equipped:
Changes to attack tab, then does nothing

Case special weapon equipped and orb visible:
Clicks special attack orb

Case special weapon equipped and orb hidden:
Changes to attack tab and clicks special

PR #2067, opened 2019-06-25
Remove costs from LocalPath

tanlines:
The original `getCosts` seemed to be trying to increase the cost of tiles around obstacles and boundaries. This is a bad idea since it would make paths avoid walking next to walls.
HOWEVER
The original `getCosts` was broken and would add random values for every tile.

Removing the costs altogether made the method stable and also increased speed significantly, see [difference in paths](https://imgur.com/a/GpYYxjs). Since this is used for calculating distance, it should not be randomised.

The previous map offset also always returned 0, the working method is `ctx.game.mapOffset()`.
`getNode` was updated in light of this change. The two classes that call `getNode` already get the offset from `ctx.game.mapOffset()`.

### DIFF
--- a/src/main/java/org/powerbot/script/rt4/LocalPath.java
+++ b/src/main/java/org/powerbot/script/rt4/LocalPath.java
@@ -37,48 +37,12 @@ public class LocalPath extends Path {
 		final int floor = client.getFloor();
 		final CollisionMap[] maps = client.getCollisionMaps();
 		final CollisionMap map;
-		if (maps == null || floor < 0 || floor >= maps.length || (map = maps[floor]) == null) {
+		final Tile base = ctx.game.mapOffset();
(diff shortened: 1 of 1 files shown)


Smileynator:
The point of the random cost, might have been to encourage behaviour that is not consistent every time (aka: anti measures for banning) which might be undesired to be removed. I can imagine there is a big gain to be made between 1ms and 70ms though.

tanlines:
@Smileynator Should be handled on a script by script basis. No point randomising all the points between you and your destination when you're just going to click your destination and let the game path find itself.

Smileynator:
Fair point :+1:

ChrisBPB:
This is done because if you click around many obstacles your waypoint market will be reset which causes odd behaviour in scripts

PR #2072, opened 2019-07-11
Initial Commit HillGiantKiller V 1.0.0

justinwagoner3:


ChrisBPB:
The parent package must be your powerbot username (e.g. coma/scripts/Script.java)

ChrisBPB:
Your file structure is correct but you have not packaged like this

PR #2088, opened 2019-10-12
Fix isValid cheack, added npc reachable check.

usrstu:


AshGreh:
Just saw your PR, good job. Could you also convert it to a lambda?

usrstu:
> Just saw your PR, good job. Could you also convert it to a lambda?

Thanks, is that possible with tile matrix?

AshGreh:
Yeah, the npc will still exhibit the same properties when executed through a lambda.
It would look something like this,
```java
private boolean isValid() {
		return ctx.input.blocking() && !ctx.properties.getProperty("randomevents.disable", "").equals("true")
				&& !ctx.npcs.select().within(5d).action("Dismiss").select(npc -> npc.interacting().equals(ctx.players.local()) && npc.tile().matrix(ctx).reachable()).isEmpty();
	}
```

usrstu:
Oh sorry didn’t know what you were asking, thought you meant convert the statement I added into a lambda.

AshGreh:
No problem. I'm just waiting on your change, once you make the changes, I will merge your request.

Strikegently:
Is the `reachable()` call necessary? Is it even possible for a random event to appear for you in a non-reachable tile?

usrstu:
> Is the `reachable()` call necessary? Is it even possible for a random event to appear for you in a non-reachable tile?

Yes I had a edge case, I was using gate but as I was using gate a event popped up. Event Npc was on the other side of gate and the bot got stuck dismissing the npc with a chat box spammed with "I cant reach that".

AshGreh:
@Strikegently agility course is another example

AshGreh:
@usrstu one last thing. Your code is not in the project's code format. The spacing doesn't match up with the standard of the project.

AshGreh:
@usrstu hey, sorry for the delayed reply. The code still doesn't follow the project's standard. @Strikegently which option in intellij do you use to reformat the code to match the rest of the project?

Strikegently:
Highlight the code block then Cltr + Alt + L

ArtificialPB:
Getting caught up in trying to enforce strict following of project's standards is the quickest way to discourage open source contributions. I'm not saying you shouldn't require all contributions to have some reasonable level of compliance with the standard, but nitpicking over formatting and such is a complete waste of time for both parties. Reformatting while doing a merge commit (via IDE) literally takes a second. Anyways, just my two cents.

AshGreh:
@ArtificialPB Agreed. I was wondering if it would be better to accept the PR and then make another PR to fix the formatting. What are your thoughts on the matter?

ArtificialPB:
@AshGreh Personally, I would pull the branch locally, add another commit to address any discrepancies from projects standards, and do a merge commit (`--no-ff` option). By opening another PR seems a bit redundant + you lose a direct, logical connection to the offending commits and clutter git history.

PR #2109, opened 2019-12-14
Add speed() function. Change speed(int) function to match JavaDocs

kThisIsCvpv:
The header for the function bamboozled me.
Sending speed(0) actually does change the speed instead of returning nothing.


kThisIsCvpv:
Also added a function that permits slowing down the default mouse (speed > 100).

AshGreh:
@kThisIsCvpv Hey Nagi, I'm planning on going with this https://github.com/powerbot/RSBot-API/blob/15137fc911aa890473862c4aa22fa323d98b08af/src/main/java/org/powerbot/script/Input.java as I had tested it.

kThisIsCvpv:
You do you. Just pls remove the confusing JavaDoc that states 0 -> return speed, when it doesn't.

PR #2143, opened 2020-04-29
removed redundant code

gutizia:
the required component data are loaded in regardless of the component being visible. With previous code ring and quiver items would return nil if you weren't looking at the component even thought they had valid data


AshGreh:
The components are loaded on login even though the equipment tab is not open. I have no objections to the PR.

PR #2164, opened 2020-06-15
Revert "Allow npc and object interactions to rely on cache models"

Strikegently:
Reverts powerbot/RSBot-API#2160


Strikegently:
@Paris emergency release - big breaking changes resolved.

PR #2165, opened 2020-06-15
Model interaction improvements

k0na:
These changes resolve the model issues that have been raised.

**Fix valid cache models yielding invalid points edge case**
No longer assume points retrieved from valid cache models are correct;
only return if not (-1,-1) and otherwise use bounding model as fallback.

**Support scaling of NPC models**
Fits models that are reused across multiple npcs to size, for example,
cow and cow calves share the same model but have different interactable
bounds.

### DIFF
--- a/src/main/java/org/powerbot/script/rt4/Actor.java
+++ b/src/main/java/org/powerbot/script/rt4/Actor.java
@@ -11,7 +11,7 @@ import java.awt.*;
  * Actor
  * A base class of all characters within Runescape.
  */
-public abstract class Actor extends Interactive implements InteractiveEntity, Nameable, Validatable , Modelable {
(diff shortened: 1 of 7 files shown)


ChrisBPB:
please don't merge this before .18 is release and testing is done

k0na:
Agreed @ChrisBPB, I've marked the pull request as a draft for that exact reason

k0na:
The changes work fine on my computer but I would appreciate if someone could git checkout the `model-interactions` branch and verify on their end before we merge this.

Paris:
Hey @k0na could you reserve the merge conflicts when you have the time.

Strikegently:
Everything seems good. I'm going to merge this. I have a change I'd like to see, but it's in the client code you can't access so I can make it.

PR #2174, opened 2020-08-12
Identify walk sequences

k0na:


Strikegently:
The updater does a bunch of stuff by reflection - I'm not positive if this will cause issues. Keep an eye on it and let me know.

PR #2176, opened 2020-08-20
Injection

constt:
TODO:

Callback for in game messages


Strikegently:
@Paris 
These are massive changes so please do NOT merge this right now.
Let's do a good amount of beta testing off of the `injection` branch with script writers before releasing it public.

PR #2178, opened 2020-09-01
Replace anonymous classes with lambda expressions

rs-scriptkitty:
As the title suggests, this pull request replaces each anonymous class that extends `Condition.Check` with a corresponding lambda expression, and therefore reduces the amount of visual clutter in the code. 😊


rs-scriptkitty:
The second commit (see 07d4749) removes the private inner class `Worlds.ClientStateCondition` (which is a concrete subclass of `Condition.Check`) in favour of using lambda expressions.

